PR #5005, opened 2026-05-21
save/restore cm->lcr_params per stream in multistream

### LINKED ISSUE #5014: LCR inconsistency with multistream
**Background**

The code maintains two structures related to the last activated LCR parameters, `cm->lcr_params` and `pbi->active_lcr`.

`cm->lcr_params` is a structure that contains the values of the LCR when it was activated.

`pbi->active_lcr` is a pointer.

These are setup in `activate_layer_configuration_record`:

```
static void activate_layer_configuration_record(AV2Decoder *pbi,
                                                int seq_lcr_id) {
  AV2_COMMON *const cm = &pbi->common;
  LayerConfigurationRecord *lcr =
      find_active_lcr(pbi, cm->xlayer_id, seq_lcr_id);
  if (lcr != NULL) {
    ...
    pbi->active_lcr = lcr;
    cm->lcr_params = *pbi->active_lcr;
```

The concern is that with multistream, only one of these is saved and restored:

```
   pbi->stream_info[stream_idx].active_lcr_buf = pbi->active_lcr;
```

but the other is used when retrieving layer information:

```
static INLINE const EmbeddedLayerInfo *get_active_embedded_layer_info(
    const AV2_COMMON *cm) {
  const LayerConfigurationRecord *lcr = &cm->lcr_params;
```

**Problem**

The problem is if you make a multistream out of two streams that contain local LCR OBUs, the checks, e.g. in check_lcr_frame_size_conformance, are sometimes applied to one stream based on the LCR from the other stream.

This causes the reference decoder to flag errors for valid bitstreams.

**Suggested fix**

Apply the LCR checks in a stream based on the LCR in the same stream.

**Note**

Manually migrated from Gitlab issue: https://gitlab.com/AOMediaCodec/avm/-/work_items/1512

SMPaluri:
Closes #5014

### DIFF
--- a/av2/decoder/decoder.h
+++ b/av2/decoder/decoder.h
@@ -298,6 +298,8 @@ typedef struct {
   struct OperatingPointSet ops_list_buf[MAX_NUM_OPS_ID];
   int ops_counter_buf;
   struct LayerConfigurationRecord *active_lcr_buf;
+  struct LayerConfigurationRecord lcr_params_buf;
+  struct LayerConfigurationRecord global_lcr_params_buf;
   struct AtlasSegmentInfo *active_atlas_segment_info_buf;
   struct quantization_matrix_set qm_list_buf[NUM_CUSTOM_QMS];
   int qm_protected_buf[NUM_CUSTOM_QMS];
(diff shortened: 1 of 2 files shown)


urvangjoshi:
@SMPaluri : Thanks for the PR.  I'll let Peter review the fix,  but keeping the PR as draft for now,  until we can resume merging of PRs.

peterderivaz (approved):
Looks good to me, thanks!

urvangjoshi (approved):

PR #5006, opened 2026-05-22
Add a missing "*/" to av2/common/annexA.c

wantehchang:
A "/*" is missing a corresponding "*/". Also indent the body of the comment block by one space to match the style of other similar comments in the file.


av2/common/annexA.c
```diff
@@ -273,22 +272,21 @@ int av2_check_profile_interop_conformance(
...
+ * ------------------------------------------------------------------------------------------------------------------
+ * 0, 1, 2                                     | 0                    | 15                   | 1.0                  |
+ * ------------------------------------------------------------------------------------------------------------------
+ * 3                                           | 1                    | 20                   | 1.667                |
+ * ------------------------------------------------------------------------------------------------------------------
+ * 4                                           | 2                    | 30                   | 2.5                  |
+ * ------------------------------------------------------------------------------------------------------------------
+ * 31                                          | -                    | -                    | -                    |
+ * ------------------------------------------------------------------------------------------------------------------
+ */
```
wantehchang, line 286: Line 286 is the most important change. The addition of "*/" fixes the following compiler warning:

```
av2/common/annexA.c:288:1: warning: '/*' within block comment [-Wcomment]
  288 | /* clang-format on */
      | ^
```

urvangjoshi (approved):

av2/common/annexA.c
wantehchang: I have only seen this warning when I build AVM as a dependency of libavif, and only during some rebuilds. (I don't remember whether this is a GCC or Clang warning.) I don't know why the compiler doesn't always warn about this.

PR #5010, opened 2026-05-26
Output value of avm_uleb_decode() is <= UINT32_MAX

wantehchang:
If the leb128() value read from the bitstream is > UINT32_MAX, avm_uleb_decode() fails and returns -1. Therefore after a successful avm_uleb_decode() call, the output value must be <= UINT32_MAX.

### DIFF
--- a/av2/decoder/obu.c
+++ b/av2/decoder/obu.c
@@ -1324,11 +1324,11 @@ static size_t read_metadata_unit_header(AV2Decoder *pbi, const uint8_t *data,
   AV2_COMMON *const cm = &pbi->common;
   size_t type_length;
   uint64_t type_value;
-  if (avm_uleb_decode(data, sz, &type_value, &type_length) < 0 ||
-      type_value > UINT32_MAX) {
+  if (avm_uleb_decode(data, sz, &type_value, &type_length) < 0) {
     cm->error.error_code = AVM_CODEC_CORRUPT_FRAME;
     return 0;
   }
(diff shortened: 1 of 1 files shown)


av2/decoder/obu.c
```diff
@@ -1320,11 +1320,11 @@ static size_t read_metadata_unit_header(AV2Decoder *pbi, const uint8_t *data,
   AV2_COMMON *const cm = &pbi->common;
   size_t type_length;
   uint64_t type_value;
-  if (avm_uleb_decode(data, sz, &type_value, &type_length) < 0 ||
-      type_value > UINT32_MAX) {
+  if (avm_uleb_decode(data, sz, &type_value, &type_length) < 0) {
     cm->error.error_code = AVM_CODEC_CORRUPT_FRAME;
     return 0;
   }
+  assert(type_value <= UINT32_MAX);
```
wantehchang, line 1331: The following code in `avm_uleb_decode()` ensures that `type_value <= UINT32_MAX`:

```
int avm_uleb_decode(const uint8_t *buffer, size_t available, uint64_t *value,
                    size_t *length) {
        ...

        // Fail on values larger than 32-bits to ensure consistent behavior on
        // 32 and 64 bit targets: value is typically used to determine buffer
        // allocation size.
        if (*value > UINT32_MAX) return -1;

        ...
}
```

urvangjoshi (approved):

PR #5016, opened 2026-05-29
Remove unreachable return statements after avm_internal_error calls. 

### LINKED ISSUE #5002: `unreachable-code-return` compile errors with LLVM 22.1.6
**What version / commit were you testing with?**

87aa8dc44e8a40f992604111a5ce83cf413ec9b7

**What steps will reproduce the problem?**

  Build AVM LLVM 22.1.6

   ```bash
   CC=${HOME}/Downloads/LLVM-22.1.6-Linux-X64/bin/clang CXX=${HOME}/Downloads/LLVM-22.1.6-Linux-X64/bin/clang++ cmake ../avm  -DENABLE_WERROR=1        && make -j100
   ```

**What is the expected output?**

No errors

**What do you see instead?**

A number of `-Wunreachable-code-return` errors:

```cpp

/tmp/avm/av2/av2_dx_iface.c:1339:14: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]                                                          
 1339 |       return cm->error.error_code;                                                                                                                                                                         
      |              ^~                                                                                                                                                                                            
/tmp/avm/av2/av2_dx_iface.c:1396:16: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]                                                          
 1396 |         return cm->error.error_code;                                                                                                                                                                       
      |                ^~                       

...

/tmp/avm/av2/decoder/decoder.c:505:12: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]                                                        
  505 |     return AVM_CODEC_ERROR;                                                                                                                                                                                
      |            ^~~~~~~~~~~~~~~                                                                                                                                                                                 
/tmp/avm/av2/decoder/decoder.c:535:12: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]                                                        
  535 |     return AVM_CODEC_ERROR;                                                                                                                                                                                
      |            ^~~~~~~~~~~~~~~                                                                                             

...


/tmp/avm/av2/av2_cx_iface.c:3681:16: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]                                                          
 3681 |         return cm->error.error_code;                                                                                                                                                                       
      |                ^~                              

....

/tmp/avm/av2/decoder/decodeframe.c:9543:13: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]                                                   
 9543 |     return -1;                                                                                                                                                                                             
      |             ^                                                                                                                                                                                              
/tmp/avm/av2/decoder/decodeframe.c:9537:13: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]                                                   
 9537 |     return -1;                                                                                                                                                                                             
      |             ^                                                                                                                                                                                              
/tmp/avm/av2/decoder/decodeframe.c:9530:13: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]                                                   
 9530 |     return -1;

....

/tmp/avm/av2/encoder/bitstream.c:6840:12: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]
 6840 |     return 0;                               
      |            ^                                
/tmp/avm/av2/encoder/bitstream.c:6831:12: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]
 6831 |     return 0;                               
      |            ^                                
/tmp/avm/av2/encoder/bitstream.c:6919:12: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]
 6919 |     return 0;                               
      |            ^                                
/tmp/avm/av2/encoder/bitstream.c:6910:12: error: 'return' will never be executed [-Werror,-Wunreachable-code-return]
 6910 |     return 0;  

```

urvangjoshi:
This fixes following type of build errors on LLVM 22.1.6:

```
error: 'return' will never be executed [-Werror,-Wunreachable-code-return]
```

Related: Ensure that setjmp is called before avm_internal_error in FilmGrainTableIOTest.

Closes #5002

### DIFF
--- a/av2/common/ccso.c
+++ b/av2/common/ccso.c
@@ -374,13 +374,11 @@ void av2_apply_ccso_filter_for_row(AV2_COMMON *cm, MACROBLOCKD *xd,
       avm_internal_error(
           &cm->error, AVM_CODEC_ERROR,
           "Invalid BRU activity in CCSO: only active SB can be filtered");
-      return;
     }
     if (cm->bridge_frame_info.is_bridge_frame) {
       avm_internal_error(
           &cm->error, AVM_CODEC_ERROR,
           "Invalid Bridge frame activity in CCSO: can not be filtered");
-      return;
     }
 
     if (cm->features.has_lossless_segment) {
@@ -533,7 +531,6 @@ void apply_ccso_filter(AV2_COMMON *cm, MACROBLOCKD *xd, int plane,
                   avm_internal_error(&cm->error, AVM_CODEC_ERROR,
                                      "Invalid BRU activity in CCSO: only "
                                      "active SB can be filtered");
-                  return;
                 }
                 if (cm->features.has_lossless_segment) {
                   ccso_filter_block_hbd_wo_buf_4x4_c(
--- a/av2/common/reconinter.c
+++ b/av2/common/reconinter.c
@@ -1972,7 +1972,6 @@ void av2_build_one_bawp_inter_predictor(
     avm_internal_error(
         (struct avm_internal_error_info *)&cm->error, AVM_CODEC_ERROR,
         "Inter BAWP template cannot outside the valid reference range");
-    return;
   } else {
     uint16_t *recon_buf = xd->plane[plane].dst.buf;
     int recon_stride = xd->plane[plane].dst.stride;
--- a/av2/common/restoration.c
+++ b/av2/common/restoration.c
@@ -1216,7 +1216,6 @@ static void pc_wiener_stripe_highbd(const RestorationUnitInfo *rui,
       avm_internal_error(
           rui->error, AVM_CODEC_ERROR,
           "Invalid BRU activity in LR: only active SB can be filtered");
-      return;
     }
     MB_MODE_INFO **mbmi_ptr_procunit = rui->mbmi_ptr + mi_offset_x;
 
@@ -1590,7 +1589,6 @@ static void wiener_nsfilter_stripe_highbd(const RestorationUnitInfo *rui,
       avm_internal_error(
           rui->error, AVM_CODEC_ERROR,
           "Invalid BRU activity in LR: only active SB can be filtered");
-      return;
     }
     MB_MODE_INFO **mbmi_ptr_procunit = rui->mbmi_ptr + mi_offset_x;
     apply_wienerns_class_id_highbd(
--- a/av2/decoder/decodeframe.c
+++ b/av2/decoder/decodeframe.c
@@ -2676,7 +2676,6 @@ static AVM_INLINE void decode_restoration_mode(AV2_COMMON *cm,
         avm_internal_error(&cm->error, AVM_CODEC_ERROR,
                            "Invalid RU size, RU size shall not be smaller than "
                            "stripe height which is 64 for 422 format");
-        return;
       }
     }
     cm->rst_info[2].restoration_unit_size =
@@ -2697,7 +2696,6 @@ static AVM_INLINE void decode_restoration_mode(AV2_COMMON *cm,
           &cm->error, AVM_CODEC_ERROR,
           "Invalid RU size, RU size shall be an integer divisor of tiles "
           "width or height, except right-most and bottom tiles");
-      return;
     }
   }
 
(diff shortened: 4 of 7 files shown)


av2/av2_cx_iface.c
```diff
@@ -3621,7 +3621,6 @@ static avm_codec_err_t ctrl_copy_reference(avm_codec_alg_priv_t *ctx,
       AV2_COMMON *cm = &ctx->cpi->common;
       avm_internal_error(&cm->error, AVM_CODEC_INVALID_PARAM,
                          "Incorrect buffer dimensions");
-      return cm->error.error_code;
```
wantehchang: avm_internal_error() calls longjmp() only when setjmp() has been called. That is not the case here.

We should also verify setjmp() has been called in the other changes in this PR.

urvangjoshi: Good point.  There seem to be a few places in this file and also `av2_dx_iface.c` where `setjmp` has not been called.
Are you suggesting to add `setjmp` to those places?

Note that the clang is reporting these error likely because of how `avm_internal_error()` function is declared:

```
#define CLANG_ANALYZER_NORETURN                                                                                                                                                      
#if defined(__has_feature)                                                                                                                                                           
#if __has_feature(attribute_analyzer_noreturn)                                                                                                                                       
#undef CLANG_ANALYZER_NORETURN                                                                                                                                                       
#define CLANG_ANALYZER_NORETURN __attribute__((analyzer_noreturn))                                                                                                                   
#endif                                                                                                                                                                               
#endif                                                                                                                                                                               
                                                                                                                                                                                     
void avm_internal_error(struct avm_internal_error_info *info,
....
```

wantehchang: In av2_cx_iface.c and av2_dx_iface.c we can delete the `avm_internal_error()` call and return the error code directly.

urvangjoshi: That makes sense.  I checked that most of the code in these files doesn't use `avm_internal_error`, so keeping only return statements instead is also consistent.
PTAL.

av2/common/ccso.c
```diff
@@ -374,13 +374,11 @@ void av2_apply_ccso_filter_for_row(AV2_COMMON *cm, MACROBLOCKD *xd,
       avm_internal_error(
           &cm->error, AVM_CODEC_ERROR,
           "Invalid BRU activity in CCSO: only active SB can be filtered");
```
wantehchang, line 376: The `av2_apply_ccso_filter_for_row()` function may be called transitively by `ccso_row_worker()`, which does not call `setjmp()`.

av2/common/reconinter.c
```diff
@@ -1972,7 +1972,6 @@ void av2_build_one_bawp_inter_predictor(
     avm_internal_error(
         (struct avm_internal_error_info *)&cm->error, AVM_CODEC_ERROR,
         "Inter BAWP template cannot outside the valid reference range");
```
wantehchang, line 1974: `av2_build_one_bawp_inter_predictor()` may be called transitively by `enc_row_mt_worker_hook()`, which does not call `setjmp()`.

Note that libaom handles this correctly. We can port libaom's changes to the AVM.

av2/av2_cx_iface.c
```diff
@@ -3619,8 +3619,6 @@ static avm_codec_err_t ctrl_copy_reference(avm_codec_alg_priv_t *ctx,
 
     if (!(frame->img.fmt & AVM_IMG_FMT_HIGHBITDEPTH)) {
       AV2_COMMON *cm = &ctx->cpi->common;
-      avm_internal_error(&cm->error, AVM_CODEC_INVALID_PARAM,
-                         "Incorrect buffer dimensions");
       return cm->error.error_code;
```
wantehchang: I suggest we move the changes to av2/av2_cx_iface.c and av2/av2_dx_iface.c to a separate PR so that they can be reviewed and merged first.

urvangjoshi: Good idea!  Moved to this PR: https://github.com/AOMediaCodec/avm/pull/5028

Marking this PR as Draft as I check about `setjmp` for other places.

PR #5027, opened 2026-06-03
avmenc.c: Misc cleanup for config->arg_ctrls

wantehchang:
Document that the current value of the ARG_CTRL_CNT_MAX macro doesn't really make sense.

Handle invalid indexes for the config->arg_ctrls array rather than assert the indexes are valid.

Use the pre-increment ++ operator on config->arg_ctrl_cnt consistently.

When adding AVME_SET_QP to the config->arg_ctrls array, look for an existing array element rather than always insert at the end of the array.

### DIFF
--- a/apps/avmenc.c
+++ b/apps/avmenc.c
@@ -447,6 +447,13 @@ void usage_exit(void) {
 }
 
 #if CONFIG_AV2_ENCODER
+// ARG_CTRL_CNT_MAX is the first dimension of the config->arg_ctrls array.
+// Note: The config->arg_ctrls array is only used for three codec controls:
+//   * AV2E_SET_TARGET_SEQ_LEVEL_IDX (may be repeated)
+//   * AVME_SET_ENABLEAUTOALTREF
+//   * AVME_SET_QP
+// So ARG_CTRL_CNT_MAX doesn't need to match the size of the av2_arg_ctrl_map
+// array.
 #define ARG_CTRL_CNT_MAX NELEMENTS(av2_arg_ctrl_map)
 #define ARG_KEY_VAL_CNT_MAX NELEMENTS(av2_key_val_args)
 #endif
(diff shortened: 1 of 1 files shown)


apps/avmenc.c
```diff
@@ -447,6 +447,13 @@ void usage_exit(void) {
 }
 
 #if CONFIG_AV2_ENCODER
+// ARG_CTRL_CNT_MAX is the first dimension of the config->arg_ctrls array.
+// Note: The config->arg_ctrls array is only used for three codec controls:
+//   * AV2E_SET_TARGET_SEQ_LEVEL_IDX (may be repeated)
+//   * AVME_SET_ENABLEAUTOALTREF
+//   * AVME_SET_QP
+// So ARG_CTRL_CNT_MAX doesn't need to match the size of the av2_arg_ctrl_map
+// array.
```
wantehchang, line 456: Urvang: If we know the maximum number of times the `--target-seq-level-idx` option can be repeated (say `N`), we can define `ARG_CTRL_CNT_MAX` as `N + 2`.

urvangjoshi (approved):

apps/avmenc.c
```diff
@@ -1200,11 +1212,21 @@ static int parse_stream_params(struct AvxEncoderConfig *global,
       const unsigned int cq_level_val = arg_parse_uint(&arg);
       const int qp_val =
           get_qindex_from_quantizer_and_warn(cq_level_val, "cq-level", "qp");
-      const int idx = config->arg_ctrl_cnt;
-      assert(idx < (int)ARG_CTRL_CNT_MAX);
```
wantehchang, line 1204: Note that it is not necessary to cast `ARG_CTRL_CNT_MAX` to `int` because the `NELEMENTS` macro already contains an `int` cast.

PR #5028, opened 2026-06-03
Remove avm_internal_error calls from av2_cx/dx_iface.c

urvangjoshi:
`setjmp` is not called at these places, so error should be returned instead.

#5002

### DIFF
--- a/av2/av2_cx_iface.c
+++ b/av2/av2_cx_iface.c
@@ -3619,8 +3619,6 @@ static avm_codec_err_t ctrl_copy_reference(avm_codec_alg_priv_t *ctx,
 
     if (!(frame->img.fmt & AVM_IMG_FMT_HIGHBITDEPTH)) {
       AV2_COMMON *cm = &ctx->cpi->common;
-      avm_internal_error(&cm->error, AVM_CODEC_INVALID_PARAM,
-                         "Incorrect buffer dimensions");
       return cm->error.error_code;
     }
     image2yuvconfig(&frame->img, &sd);
@@ -3676,8 +3674,6 @@ static avm_codec_err_t ctrl_copy_new_frame_image(avm_codec_alg_priv_t *ctx,
 
       if (!(new_img->fmt & AVM_IMG_FMT_HIGHBITDEPTH)) {
         AV2_COMMON *cm = &ctx->cpi->common;
-        avm_internal_error(&cm->error, AVM_CODEC_INVALID_PARAM,
-                           "Incorrect buffer dimensions");
         return cm->error.error_code;
       }
       image2yuvconfig(new_img, &sd);
--- a/av2/av2_cx_iface.c
+++ b/av2/av2_cx_iface.c
@@ -3618,8 +3618,7 @@ static avm_codec_err_t ctrl_copy_reference(avm_codec_alg_priv_t *ctx,
     YV12_BUFFER_CONFIG sd;
 
     if (!(frame->img.fmt & AVM_IMG_FMT_HIGHBITDEPTH)) {
-      AV2_COMMON *cm = &ctx->cpi->common;
-      return cm->error.error_code;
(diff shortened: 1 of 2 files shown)


av2/av2_cx_iface.c
```diff
@@ -3619,8 +3619,6 @@ static avm_codec_err_t ctrl_copy_reference(avm_codec_alg_priv_t *ctx,
 
     if (!(frame->img.fmt & AVM_IMG_FMT_HIGHBITDEPTH)) {
       AV2_COMMON *cm = &ctx->cpi->common;
-      avm_internal_error(&cm->error, AVM_CODEC_INVALID_PARAM,
-                         "Incorrect buffer dimensions");
       return cm->error.error_code;
```
wantehchang: IMPORTANT: Change these two lines to:

```
      ERROR("Incorrect buffer dimensions");
```

Note: The error message "Incorrect buffer dimensions") comes from https://gitlab.com/AOMediaCodec/avm/-/merge_requests/384. I think the error message should be "Incorrect buffer bit depth". You can ignore this error message issue in this PR.

av2/av2_cx_iface.c
```diff
@@ -3676,8 +3674,6 @@ static avm_codec_err_t ctrl_copy_new_frame_image(avm_codec_alg_priv_t *ctx,
 
       if (!(new_img->fmt & AVM_IMG_FMT_HIGHBITDEPTH)) {
         AV2_COMMON *cm = &ctx->cpi->common;
-        avm_internal_error(&cm->error, AVM_CODEC_INVALID_PARAM,
-                           "Incorrect buffer dimensions");
         return cm->error.error_code;
```
wantehchang: IMPORTANT: Change these two lines to:

```
      ERROR("Incorrect buffer dimensions");
```

av2/av2_dx_iface.c
```diff
@@ -1391,8 +1389,6 @@ static avm_codec_err_t ctrl_copy_new_frame_image(avm_codec_alg_priv_t *ctx,
       YV12_BUFFER_CONFIG sd;
       if (!(img->fmt & AVM_IMG_FMT_HIGHBITDEPTH)) {
         AV2_COMMON *cm = &frame_worker_data->pbi->common;
-        avm_internal_error(&cm->error, AVM_CODEC_INVALID_PARAM,
-                           "Incorrect buffer dimensions");
         return cm->error.error_code;
```
wantehchang: Change these two lines to:

```
      set_error_detail(ctx, "Incorrect buffer dimensions");
      return AVM_CODEC_INVALID_PARAM;
```

av2/av2_dx_iface.c
```diff
@@ -1334,8 +1334,6 @@ static avm_codec_err_t ctrl_copy_reference(avm_codec_alg_priv_t *ctx,
     FrameWorkerData *const frame_worker_data = (FrameWorkerData *)worker->data1;
     if (!(frame->img.fmt & AVM_IMG_FMT_HIGHBITDEPTH)) {
       AV2_COMMON *cm = &frame_worker_data->pbi->common;
-      avm_internal_error(&cm->error, AVM_CODEC_INVALID_PARAM,
-                         "Incorrect buffer dimensions");
       return cm->error.error_code;
```
wantehchang: av2/av2_dx_iface.c doesn't have the `ERROR()` macro, so we need to use a different approach.

Change these two lines to:

```
      set_error_detail(ctx, "Incorrect buffer dimensions");
      return AVM_CODEC_INVALID_PARAM;
```

av2/av2_cx_iface.c
urvangjoshi: Both good points!  Updated and fixed the error message as well (here and elsewhere).

wantehchang (approved):

PR #5034, opened 2026-06-04
Fix run-to-run non-determinism in non-separable Wiener filter

xinzhao-apple:
Fix encoder run-to-run non-determinism caused by uninitialized memory in the non-separable Wiener filter restoration search path (av2/encoder/pickrst.c).
Multiple runs of the same encoding with identical inputs produce different chroma (U/V) results while luma (Y) is always identical. Reproduced with 4:2:2 12-bit content (SolLevanteFace) at QP 135, cpu-used=0, threads=2, described in issue #1500 , and r2r issue still occurs for some inter frame occasionally before this fix.


urvangjoshi:
@jianle973 :  pls review.

jianle973 (approved):
LGTM, Thanks!

PR #5041, opened 2026-06-05
Fix RDO mode skipping corruption with reduced reference frames

kslu-aom:
When some reference frames are disabled, the encoder was making incorrect mode skipping decisions because it was using uninitialized or stale MV SAD values in x->pred_mv_sad from these disabled reference frames. This corrupted the RDO skipping logic, leading to suboptimal encoding with --max-reference-frames < 7

This fixes this issue by properly intializing the SAD values and adding a check to ensure that only enabled reference frames are considered when calculating min_pred_mv_sad.

In MultiLayerTest, TMVP is disabled because the test drops TL1 frames and causes the DPB to diverge at encoder and decoder, leading to TMVP mismatches. The test passes before because the encoder issue accidentally avoids using the TMVP, which hided the test issue. Since this is a test issue, not an encoder issue, we turn off TMVP in the test.

Coding performance with 17 frames on A4 and A5 with different --max-reference-frames

          Before   After
refs=7    +0.00    +0.00
refs=6    +0.47    +0.06
refs=5    +0.57    +0.13
refs=4    +0.61    +0.16
refs=3    +0.70    +0.24

### DIFF
--- a/av2/encoder/rdopt.c
+++ b/av2/encoder/rdopt.c
@@ -6693,8 +6693,11 @@ static AVM_INLINE void init_mode_skip_mask(mode_skip_mask_t *mask,
   int min_pred_mv_sad = INT_MAX;
   MV_REFERENCE_FRAME ref_frame;
   for (ref_frame = 0; ref_frame < cm->ref_frames_info.num_total_refs;
-       ++ref_frame)
-    min_pred_mv_sad = AVMMIN(min_pred_mv_sad, x->pred_mv_sad[ref_frame]);
+       ++ref_frame) {
+    if (cm->ref_frame_flags & (1 << ref_frame)) {
+      min_pred_mv_sad = AVMMIN(min_pred_mv_sad, x->pred_mv_sad[ref_frame]);
+    }
+  }
 
   min_pred_mv_sad = AVMMIN(min_pred_mv_sad, x->pred_mv_sad[TIP_FRAME_INDEX]);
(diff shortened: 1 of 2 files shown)


urvangjoshi:
@kslu-aom :  This test seems to be failing:

```
[ RUN      ] AV2/MultiLayerTest.MultiLayerTest2Embedded2TemporaSframe/0
../../test/multi_layer_test.cc:253: Failure
Expected equality of these values:
  AVM_CODEC_OK
    Which is: 0
  res_dec
    Which is: 7
Failed to decode tile data
```

P.S.  Finding the failing test from CI logs is tricky (due to intermingled logs).  I'll add info on this in the wiki a bit later.

urvangjoshi:
@marpan-max :  pls review the multi_layer_test update from Lester here (The 2nd commit in this PR: https://github.com/AOMediaCodec/avm/pull/5041/changes/6fd1db3e049c8a45bc554da86ac8068a2de1e884)

marpan-max (approved):

kslu-aom:
Full CTC results, 130 frames RA w/o B2

| #refs | BD-PSNR | EncTime |
|--------|--------|--------|
| 7 | +0.00 | 100% |
| 6 | +0.19 | 94% |
| 5 | +0.45 | 89% |
| 4 | +0.61 | 85% |
| 3 | +0.91 | 79% |

PR #5042, opened 2026-06-05
decoder.c: use internal_error correctly

urvangjoshi:
Migrated from libaom:
https://aomedia-review.git.corp.google.com/c/aom/+/203641

Also, a similar fix in ctrl_copy_reference().

#5044

### DIFF
--- a/av2/av2_dx_iface.c
+++ b/av2/av2_dx_iface.c
@@ -1301,7 +1301,8 @@ static avm_codec_err_t ctrl_set_reference(avm_codec_alg_priv_t *ctx,
   av2_ref_frame_t *const data = va_arg(args, av2_ref_frame_t *);
 
   if (data) {
-    avm_image_t *hbd_img = NULL;
+    // Must be declared as `volatile` due to the `setjmp` call below.
+    avm_image_t *volatile hbd_img = NULL;
     av2_ref_frame_t *const frame = data;
     YV12_BUFFER_CONFIG sd;
     AVxWorker *const worker = ctx->frame_worker;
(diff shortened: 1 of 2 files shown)


av2/av2_dx_iface.c
```diff
@@ -1337,7 +1347,18 @@ static avm_codec_err_t ctrl_copy_reference(avm_codec_alg_priv_t *ctx,
       return AVM_CODEC_INVALID_PARAM;
     }
     image2yuvconfig(&frame->img, &sd);
-    return av2_copy_reference_dec(frame_worker_data->pbi, frame->idx, &sd);
+    avm_codec_err_t res = AVM_CODEC_OK;
+    struct avm_internal_error_info *const error =
+        &frame_worker_data->pbi->common.error;
+    if (setjmp(error->jmp)) {
+      error->setjmp = 0;
```
wantehchang, line 1353: error->setjmp = 0; is in both the if and else branches. Move it after the if-else statement.

Note: This function doesn't need to do any cleanup before returning, so we can use the more commonly-used control structure like the one used in the [libaom CL.](https://aomedia-review.googlesource.com/c/aom/+/203641)

av2/av2_dx_iface.c
```diff
@@ -1315,9 +1316,18 @@ static avm_codec_err_t ctrl_set_reference(avm_codec_alg_priv_t *ctx,
...
       image2yuvconfig(&frame->img, &sd);
     }
-    avm_codec_err_t res =
-        av2_set_reference_dec(&frame_worker_data->pbi->common, frame->idx,
-                              frame->use_external_ref, &sd);
+    avm_codec_err_t res = AVM_CODEC_OK;
+    struct avm_internal_error_info *const error =
+        &frame_worker_data->pbi->common.error;
+    if (setjmp(error->jmp)) {
+      error->setjmp = 0;
```
wantehchang: `error->setjmp = 0;` is in both the `if` and `else` branches. Move it after the if-else statement.

wantehchang (approved):
cc: @jeromejj @jzern

av2/av2_dx_iface.c
urvangjoshi: Thanks!  I changed this to use libaom-like control structure.

av2/av2_dx_iface.c
urvangjoshi: Done.

wantehchang (approved):

PR #5048, opened 2026-06-08
Fix encoder crash at low (negative) QP for trellis related function

xinzhao-apple:
Two encoder-only trellis bugs:

1. Clamp base_ctx/mid_ctx to LF_SIG_COEF_CONTEXTS-2 / LF_LEVEL_CONTEXTS-2 in C and AVX2. base_diag_ctx + (coef & 0xF) could reach 47, exceeding the 33-entry context table. The -2 margin keeps the AVX2 [ctx][1] 8-byte loads in bounds.

2. av2_pre_quant clamped qIdx to (1<<16)-1, but prequant_t::qIdx is int16_t — values >= 32768 sign-truncate to negative and produce wild pointers downstream. Tighten the clamp to INT16_MAX in C and AVX2.

Decoder unaffected; bitstreams remain spec-compliant.

I think this bugfix should impact also 8-bit and 10-bit very low QP (negative QPs), but so far I only verified on 12bit sequences with negative QP setting.

Issue can bre reproduced: Waterfall / ParkJoy / NocturneDance 3840x2160 yuv422p12 at qp=-41 or qp=-16.

### DIFF
--- a/av2/encoder/trellis_quant.c
+++ b/av2/encoder/trellis_quant.c
@@ -396,7 +396,7 @@ void av2_pre_quant_c(tran_low_t tqc, struct prequant_t *pqData,
   tran_low_t abs_tqc = abs(tqc);
 
   tran_low_t qIdx = (int)AVMMAX(
-      1, AVMMIN(((1 << 16) - 1),
+      1, AVMMIN(INT16_MAX,
                 ((int64_t)abs_tqc * quant_ptr[scan_pos != 0] + add) >> shift));
   pqData->qIdx = qIdx;
 
--- a/av2/encoder/x86/trellis_quant_avx2.c
+++ b/av2/encoder/x86/trellis_quant_avx2.c
@@ -200,7 +200,7 @@ void av2_pre_quant_avx2(tran_low_t tqc, struct prequant_t *pqData,
   int32_t abs_tqc = abs(tqc);
 
   int32_t qIdx = (int)AVMMAX(
-      1, AVMMIN(((1 << 16) - 1),
+      1, AVMMIN(INT16_MAX,
                 ((int64_t)abs_tqc * quant_ptr[scan_pos != 0] + add) >> shift));
   pqData->qIdx = qIdx;
 


urvangjoshi:
@joeyoung-yt :  I'll wait for your review.

av2/encoder/trellis_quant.c
```diff
@@ -658,6 +662,12 @@ void av2_get_rate_dist_lf_luma_c(const struct tcq_param_t *p,
     int base_ctx =
         get_base_diag_ctx(diag_ctx) + get_base_ctx(coeff_ctx->coef[i]);
     int mid_ctx = get_mid_diag_ctx(diag_ctx) + get_mid_ctx(coeff_ctx->coef[i]);
+    // Saturate combined contexts to avoid out-of-bounds reads.
+    // base_diag_ctx ∈ [0, 32] and get_base_ctx(...) ∈ [0, 15] can combine to
+    // 47, exceeding LF_SIG_COEF_CONTEXTS = 33.
+    if (base_ctx > LF_SIG_COEF_CONTEXTS - 2)
```
joeyoung-yt: Are these checks triggered even with the above clamping (1 << 15) - 1?

av2/encoder/x86/trellis_quant_avx2.c
```diff
@@ -199,8 +199,11 @@ void av2_pre_quant_avx2(tran_low_t tqc, struct prequant_t *pqData,
   int32_t add = -((2 << shift) >> 1);
   int32_t abs_tqc = abs(tqc);
 
+  // FIX: cap qIdx to (1<<15) - 1 = INT16_MAX. pqData->qIdx is int16_t; the
```
urvangjoshi: One suggestions:
- You can remove the comment from code and add this explanation in PR description instead.

Also, some questions:
- Given your explanation, why not use `INT16_MAX` directly instead of `((1 << 15) - 1)`?
- Your comment says `pqData->qIdx is int16_t`.  If so, why is `qIdx` declared at `int32_t`?
- Also, why the `int` cast at line 205, and not `int16_t` cast?

av2/encoder/trellis_quant.c
```diff
@@ -395,8 +395,12 @@ void av2_pre_quant_c(tran_low_t tqc, struct prequant_t *pqData,
   tran_low_t add = -((2 << shift) >> 1);
   tran_low_t abs_tqc = abs(tqc);
 
+  // FIX: cap qIdx to (1<<15) - 1 = 32767 = INT16_MAX, not (1<<16) - 1 = 65535.
```
urvangjoshi: Same here

av2/encoder/x86/trellis_quant_avx2.c
xinzhao-apple: ``You can remove the comment from code and add this explanation in PR description instead.``

Yes, I have removed the comments in the code, the current PR description already covered the comment, so I did not make further edit on the PR description.

``Given your explanation, why not use INT16_MAX directly instead of ((1 << 15) - 1)?``

I have cleaned up the code using INT16_MAX. Thanks for the suggestion.

``Your comment says pqData->qIdx is int16_t. If so, why is qIdx declared at int32_t?``

The local qIdx is tran_low_t (= int32_t) because the inner expression ((int64_t)abs_tqc * quant_ptr[...] + add) >> shift can exceed int16_t range before the AVMMIN clamp, that's the very value we're clamping. After the clamp the result fits in [1, INT16_MAX] and is safely stored into pqData->qIdx. The wider local type was pre-existing on main; this PR didn't touch it.

``Also, why the int cast at line 205, and not int16_t cast?``

The (int) cast is pre-existing on main branch, this PR didn't introduce it. Since the clamped value fits in [1, INT16_MAX], an int16_t cast would be technically more accurate, but (int) is harmless. I can add additional code changes to tighten in a follow-up if it looks better to you, in this PR I am trying to make minimal changes with the issue fixed.

av2/encoder/trellis_quant.c
xinzhao-apple: Removed these code comments in the updated PR.

av2/encoder/trellis_quant.c
xinzhao-apple: Thanks Joe for the comment. Agreed that the base_ctx checks aren't necessary. I've reverted those checks in the latest commit. I also re-ran Waterfall 4K 4:2:2 12-bit at qp=-16, and the encoder crash at the first frame no longer reproduces.

av2/encoder/x86/trellis_quant_avx2.c
joeyoung-yt: Looking at the code, it seems that qIdx could be 32 bits (and there seems to be no SIMD code relying on it being 16 bits). If it were upgraded to 32 bits, then the clipping value could be larger (say, 1 << 19).
Maybe add this as a feature request?

av2/encoder/x86/trellis_quant_avx2.c
```diff
@@ -720,6 +720,8 @@ void av2_get_rate_dist_lf_luma_avx2(const struct tcq_param_t *p,
     mid_ctx = _mm256_srli_epi16(mid_ctx, 4);
     __m256i mid_diag = _mm256_set1_epi16(mid_diag_ctx);
     mid_ctx = _mm256_add_epi16(mid_ctx, mid_diag);
+    __m256i mid_ctx_max = _mm256_set1_epi16((int16_t)(LF_LEVEL_CONTEXTS - 2));
```
joeyoung-yt: This can also be removed?

av2/encoder/trellis_quant.c
```diff
@@ -658,6 +658,7 @@ void av2_get_rate_dist_lf_luma_c(const struct tcq_param_t *p,
     int base_ctx =
         get_base_diag_ctx(diag_ctx) + get_base_ctx(coeff_ctx->coef[i]);
     int mid_ctx = get_mid_diag_ctx(diag_ctx) + get_mid_ctx(coeff_ctx->coef[i]);
+    if (mid_ctx > LF_LEVEL_CONTEXTS - 2) mid_ctx = LF_LEVEL_CONTEXTS - 2;
```
joeyoung-yt: This can also be removed?

av2/encoder/x86/trellis_quant_avx2.c
xinzhao-apple: @joeyoung-yt I agree with your suggestion to change qIdx from 16-bit to 32-bit. However, would it be OK if we proceed with this MR for now since it resolves the encoder crash, and consider the data-type change as a separate MR?

joeyoung-yt: Yes, that makes sense to me.
Thanks for finding this fix!

av2/encoder/x86/trellis_quant_avx2.c
xinzhao-apple: After further checking, I think you are right, the following line is accessing cost_low_tbl[idx][14][..][..], which is out of bound, 

``__m128i rate_67 = _mm_loadu_si64(&cost_low_tbl[idx][ctx3][1]);``

but those out-of-bound values are not used in the calculation after 

``__m128i rate_4567 = _mm_unpacklo_epi32(rate_45, rate_67);``

I have updated the PR accordingly. Now the code changes is only the qIdx.

av2/encoder/trellis_quant.c
xinzhao-apple: Updated the PR by removing this change. Thanks for review!

joeyoung-yt (approved):
LGTM

Issue #5044, opened 2026-06-05
Ensure that setjmp() is called in the code paths that call avm_internal_error()


urvangjoshi:
### Describe the requested feature / task
This is a pre-requisite task for removing redundant `return` statements immediately following `avm_internal_error()` calls.

### Describe how the feature will be implemented / how the task will be completed
Some fixes can be migrated from libaom:
e.g.

- [x] Decoder API fixes from here: https://aomedia-review.git.corp.google.com/c/aom/+/203641
- [ ] Multi-threading related fixes starting from here: https://aomedia.issues.chromium.org/issues/42302297#comment58

But there maybe other fixes also required: TBD based on further analysis of the code.

urvangjoshi:
**Regarding  porting of multi-thread related fixes from libaom**:
-  I have ported 4 PRs so far.
- At this point,  we need to port PRs starting from one in this comment https://aomedia.g-issues.chromium.org/issues/42302297#comment63  and ones after that.
However, as noticed by @wantehchang , some of these may require porting other dependent changes -- or adjusting the PRs to work correctly with AVM, due to differences in the codebase.

**@yunqingwang1 :  Ittiam folks who worked on this for libaom maybe best to do this work.  Could you see if they can take up porting these remaining fixes?**

Note: after all the fixes have been ported,  my draft PR https://github.com/AOMediaCodec/avm/pull/5016 will need to be rebased / updated to ensure all return statements after `avm_internal_error` have been removed.

PR #5050, opened 2026-06-08
Fix sync_enc_workers to read error status of master thread

urvangjoshi:
In sync_enc_workers(), winterface->sync() is invoked on the worker threads and the same is used to read the error status of each thread. The sync of master thread was avoided to facilitate FPMT, but its error status still needs to be read to facilitates the error propagation. This CL initializes 'had_error' with the error status of the master thread to fix this issue.

Migrated from libaom:
https://aomedia-review.git.corp.google.com/c/aom/+/176341

#5044


urvangjoshi:
@wantehchang :  I have a bunch of such fixes that I plan to send for review one by one.  Let me know if a single PR with multiple commits is better.

wantehchang (approved):

wantehchang:
Urvang: I prefer porting one commit at a time. Thanks!

PR #5052, opened 2026-06-09
Add signaling cost of skip mode and inter-intra to intrabc

yeqing-wu:
When computing the signaling rate cost for intrabc mode, the code does 
not include the cost of the skip-mode and inter-intra flags, both of 
which are signaled before the intrabc flag. As a result, the rate cost 
of intrabc is under-counted.

Under the 65-frame CWG CTC testing condition:
- RA B2: −0.21% PSNR-YUV BD-rate (gain)
- LD B2: −0.45% PSNR-YUV BD-rate (gain)

No encoding time increase.


urvangjoshi:
@msarwer2 :  could you review the IntraBC related changes?

msarwer2:
Looks good to me @urvangjoshi @yeqing-wu

urvangjoshi (approved):

PR #5056, opened 2026-06-09
Propagate error from worker to main thread during tile-MT

urvangjoshi:
This change sets a longjmp target in enc_worker_hook() to propagate error from worker threads to the main thread during tile multi-threading. Also, derives the error_info inside sync_enc_workers() to propagate the exact error details further.

Migrated from libaom:
https://aomedia-review.git.corp.google.com/c/aom/+/176342

#5044

### DIFF
--- a/av2/av2_cx_iface.c
+++ b/av2/av2_cx_iface.c
@@ -3378,7 +3378,9 @@ static avm_codec_err_t encoder_encode(avm_codec_alg_priv_t *ctx,
           &dst_end_time_stamp_la, !img, timestamp_ratio);
       if (status != -1) {
         if (status != AVM_CODEC_OK) {
-          avm_internal_error(&cpi_lap->common.error, AVM_CODEC_ERROR, NULL);
+          avm_internal_error(&cpi_lap->common.error,
+                             cpi->common.error.error_code, "%s",
+                             cpi->common.error.detail);
         }
         cpi_lap->seq_params_locked = 1;
       }
@@ -3412,7 +3414,8 @@ static avm_codec_err_t encoder_encode(avm_codec_alg_priv_t *ctx,
       cx_time += avm_usec_timer_elapsed(&timer);
       if (status == -1) break;
       if (status != AVM_CODEC_OK) {
-        avm_internal_error(&cpi->common.error, AVM_CODEC_ERROR, NULL);
+        avm_internal_error(&cpi->common.error, cpi->common.error.error_code,
+                           "%s", cpi->common.error.detail);
       }
 
       const int mlayer_id = cpi->common.mlayer_id;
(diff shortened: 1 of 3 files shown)


av2/av2_cx_iface.c
```diff
@@ -3412,7 +3414,8 @@ static avm_codec_err_t encoder_encode(avm_codec_alg_priv_t *ctx,
       cx_time += avm_usec_timer_elapsed(&timer);
       if (status == -1) break;
       if (status != AVM_CODEC_OK) {
-        avm_internal_error(&cpi->common.error, AVM_CODEC_ERROR, NULL);
+        avm_internal_error(&cpi->common.error, cpi->common.error.error_code,
+                           "%s", cpi->common.error.detail);
```
wantehchang: This change seems incorrect. I think we should pass `status` as the error code, with no error message:

```
        avm_internal_error(&cpi->common.error, status, NULL);
```

Suppose `cpi->common.error.detail` is set. Then `av2_get_compressed_data()` must have called `avm_internal_error()`, and we would not have reached here -- a contradiction! Therefore, `cpi->common.error.detail` is not set.

Another problem is that `cpi->common.error.detail` can only be used when `cpi->common.error.has_detail` is true.

av2/av2_cx_iface.c
```diff
@@ -3378,7 +3378,9 @@ static avm_codec_err_t encoder_encode(avm_codec_alg_priv_t *ctx,
           &dst_end_time_stamp_la, !img, timestamp_ratio);
       if (status != -1) {
         if (status != AVM_CODEC_OK) {
-          avm_internal_error(&cpi_lap->common.error, AVM_CODEC_ERROR, NULL);
+          avm_internal_error(&cpi_lap->common.error,
+                             cpi->common.error.error_code, "%s",
+                             cpi->common.error.detail);
```
wantehchang: Please see my comment at line 3418 below. This change has both of the problems I mentioned in that comment, plus an additional problem: since we pass `cpi_lap` to the av2_get_compressed_data()` call, we should be using `cpi_lap`, not `cpi`.

I think we should just pass `status` to the `av2_internal_error()` call with no error message:

```
          avm_internal_error(&cpi_lap->common.error, status, NULL);
```

wantehchang (approved):
cc: @jzern

LGTM with suggested changes. James, could you check my comments in av2/av2_cx_iface.c? Thanks.

av2/av2_cx_iface.c
urvangjoshi: Done.

av2/av2_cx_iface.c
urvangjoshi: Done.

jzern (approved):

PR #5057, opened 2026-06-09
encoder_encode: clear setjmp flags in cpi, cpi_lap

wantehchang:
After we call setjmp() on cpi and cpi_lap (if cpi_lap is not null), we need to clear the setjmp flags in both cpi and cpi_lap (if cpi_lap is not null) before returning.

#5044

### DIFF
--- a/av2/av2_cx_iface.c
+++ b/av2/av2_cx_iface.c
@@ -3210,6 +3210,9 @@ static avm_codec_err_t encoder_encode(avm_codec_alg_priv_t *ctx,
   // before it returns.
   if (setjmp(cpi->common.error.jmp)) {
     cpi->common.error.setjmp = 0;
+    if (cpi_lap != NULL) {
+      cpi_lap->common.error.setjmp = 0;
+    }
     res = update_error_state(ctx, &cpi->common.error);
     avm_clear_system_state();
     return res;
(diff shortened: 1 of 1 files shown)


av2/av2_cx_iface.c
```diff
@@ -3210,6 +3210,9 @@ static avm_codec_err_t encoder_encode(avm_codec_alg_priv_t *ctx,
   // before it returns.
   if (setjmp(cpi->common.error.jmp)) {
     cpi->common.error.setjmp = 0;
+    if (cpi_lap != NULL) {
+      cpi_lap->common.error.setjmp = 0;
+    }
```
wantehchang, line 3215: James, Urvang: Please review this PR carefully. I noticed this bug while reviewing Urvang's PR https://github.com/AOMediaCodec/avm/pull/5056.

av2/av2_cx_iface.c
```diff
@@ -3284,7 +3288,13 @@ static avm_codec_err_t encoder_encode(avm_codec_alg_priv_t *ctx,
...
         hbd_img = avm_img_alloc(NULL, img->fmt | AVM_IMG_FMT_HIGHBITDEPTH,
                                 img->d_w, img->d_h, 32);
-        if (!hbd_img) return AVM_CODEC_MEM_ERROR;
+        if (!hbd_img) {
+          cpi->common.error.setjmp = 0;
+          if (cpi_lap != NULL) {
+            cpi_lap->common.error.setjmp = 0;
+          }
+          return AVM_CODEC_MEM_ERROR;
+        }
```
wantehchang, line 3297: This duplicates the code at the end of the function. Alternatively, we can add a label `cleanup` right before line 3593, and change this if statement to:

```
        if (!hbd_img) {
          res = AVM_CODEC_MEM_ERROR;
          goto cleanup;
        }
```

Would you prefer using goto instead?

jzern: This is OK, though there are a few gotos around, mostly in the examples I think.

jzern (approved):

urvangjoshi (approved):

PR #5067, opened 2026-06-11
Fixed an error that occurred when compiling the executable file (#5066)

snowman20230101:


snowman20230101:
This fixes app-only builds where ENABLE_APPS=ON but ENABLE_TESTS,
ENABLE_EXAMPLES, and ENABLE_TOOLS are OFF. avmdec/avmenc still need
libyuv/webm include paths and object sources.

Verified locally:
cmake --build build_fast --target avmdec -j8
cmake --build build_fast --target avmenc -j8

PR #5096, opened 2026-07-13
Invoke av2_setup_ref_frame_sides() unconditionally

satheeshkumar02:
In the reference commit, `cm->ref_frame_side[]` and `cm->ref_frame_relative_dist[]` were computed in two separate paths: either inside `av2_setup_motion_field()` when `allow_ref_frame_mvs` is true, or in `av2_setup_ref_frame_sides()`. In this patch, `av2_setup_ref_frame_sides()` is called unconditionally to populate these reference frame arrays, thus removing the duplicate computations inside `av2_setup_motion_field()`.

### DIFF
--- a/av2/common/mvref_common.c
+++ b/av2/common/mvref_common.c
@@ -3993,8 +3993,6 @@ static void fill_id_offset_sample_gap(AV2_COMMON *cm) {
 void av2_setup_motion_field(AV2_COMMON *cm) {
   const OrderHintInfo *const order_hint_info = &cm->seq_params.order_hint_info;
 
-  memset(cm->ref_frame_side, 0, sizeof(cm->ref_frame_side));
-  memset(cm->ref_frame_relative_dist, 0, sizeof(cm->ref_frame_relative_dist));
   if (order_hint_info->order_hint_bits_minus_1 < 0) return;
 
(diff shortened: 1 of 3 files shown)


satheeshkumar02:
@yunqingwang1 @urvangjoshi Please review this refactoring change.
/cc @harishdm @ranjit-tulabandu @cherma-ittiam @deepa-kg @RemyaPrakasan

urvangjoshi:
@satheeshkumar-ittiam :  can you explain the rationale behind this change?   And why are we able to remove some code with this change? 
(Add this info to PR description).

satheeshkumar02:
> @satheeshkumar-ittiam : can you explain the rationale behind this change? And why are we able to remove some code with this change? (Add this info to PR description).

In the reference commit, `cm->ref_frame_side[]` and `cm->ref_frame_relative_dist[]` were computed in two separate paths: inside `av2_setup_motion_field()` when `allow_ref_frame_mvs` is true, and in `av2_setup_ref_frame_sides()` otherwise. By calling `av2_setup_ref_frame_sides()` unconditionally before the `allow_ref_frame_mvs` check, these reference frame arrays are always populated through one path, Hence the computations inside `av2_setup_motion_field()` can be removed.

urvangjoshi (approved):

PR #5103, opened 2026-07-13
Skip IST/STX cascade for intra-luma zero-quant primary

yeqing-wu:
Extend the intra-luma IST/STX pre-skip and post-trellis gates to cover eob == 0 in addition to the existing eob == 1 (DC-only) case. The eob == 0 branch kills candidates only when stx > 0, so a non-DCT primary with stx == 0 that quantizes to all-zero can still win via skip coding.

Add a pre-quant L-infinity gate that predicts primary FP-quant eob = 0 directly from the post-primary coefficients (max |coeff| below the FP quantizer kill threshold min(dq0, dq1) >> (4 + log_scale)). When fired, av2_quant + av2_optimize_b + cost_coeffs are skipped and the same eob_found / DCT_DCT propagation as the post-quant gate is applied. Scoped to intra Y, stx == 0, non-DC-only, IST-enabled, non-QM blocks.

Gated behind a new tx_sf.prune_intra_ist_stx_by_zero_eob flag, off by default and enabled at speed >= 1. Encoding at speed 0 is unchanged from the baseline.

STATS_CHANGED

Anchor: commit dd84ac96d1
Speed 1 (cpu-used=1): FG16 CTC (33 frames, RA)

```
  | A1         |  0.06 | -0.20 | -0.34 |  0.02 |   98 |  100 |
  | A2         |  0.09 | -0.18 | -0.12 |  0.07 |   97 |  100 |
  | A3         |  0.05 |  0.58 |  0.48 |  0.07 |   96 |   99 |
  | A4         |  0.09 |  0.22 | -3.02 | -0.01 |   97 |   99 |
  | A5         | -0.12 | -2.35 |  2.22 | -0.12 |   97 |  101 |
  | B1         |  0.14 | -0.94 | -0.28 |  0.09 |   97 |  101 |
  | B2         | -0.48 | -0.71 |  0.30 | -0.46 |   97 |   97 |
  | Avg w/o B2 |  0.07 | -0.32 | -0.24 |  0.04 |   97 |  100 |
```

### DIFF
--- a/av2/encoder/tx_search.c
+++ b/av2/encoder/tx_search.c
@@ -2094,6 +2094,29 @@ static bool prune_sec_txfm_rd_eval(int64_t sec_tx_sse_to_be_coded,
 // Encoder-only variable to reudce the search complexity of IST
 #define IST_REDUCED_SEARCH_SET_SIZE 4
 
+// Return 1 if primary FP-quant is guaranteed to produce eob = 0 across the
+// IST window, 0 otherwise. Uses an L-infinity scan of the post-primary
+// coefficients against the FP quantizer's kill threshold
+//   abs(coeff) < min(dequant[0], dequant[1]) >> (4 + log_scale)
+// which upper-bounds the per-position quantizer kill rule
+// (av2_highbd_quantize_fp_facade in av2_quantize.c).
+static INLINE int check_primary_quant_all_zero(const tran_low_t *coeff,
+                                               const int32_t dequant_QTX[2],
+                                               int width, int height,
+                                               int log_scale) {
+  const int shift = 4 + log_scale;
+  const int32_t deq_min = AVMMIN(dequant_QTX[0], dequant_QTX[1]);
+  const int32_t thr = deq_min >> shift;
+  if (thr <= 0) return 0;
+  const uint32_t uthr = (uint32_t)thr;
+  const int n = width * height;
+  for (int i = 0; i < n; ++i) {
+    const int32_t a = (coeff[i] < 0) ? -coeff[i] : coeff[i];
+    if ((uint32_t)a >= uthr) return 0;
+  }
+  return 1;
+}
+
 // Search for the best transform type for a given transform block.
 // This function can be used for both inter and intra, both luma and chroma.
 static void search_tx_type(const AV2_COMP *cpi, MACROBLOCK *x, int plane,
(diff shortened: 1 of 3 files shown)


yeqing-wu:
Hello all,

Could you please review this PR?

Thanks
Yeqing

jingninghan (approved):

av2/encoder/tx_search.c
```diff
@@ -2094,6 +2094,29 @@ static bool prune_sec_txfm_rd_eval(int64_t sec_tx_sse_to_be_coded,
 // Encoder-only variable to reudce the search complexity of IST
 #define IST_REDUCED_SEARCH_SET_SIZE 4
 
+// Return 1 if primary FP-quant is guaranteed to produce eob = 0 across the
+// IST window, 0 otherwise. Uses an L-infinity scan of the post-primary
+// coefficients against the FP quantizer's kill threshold
+//   abs(coeff) < min(dequant[0], dequant[1]) >> (4 + log_scale)
+// which upper-bounds the per-position quantizer kill rule
+// (av2_highbd_quantize_fp_facade in av2_quantize.c).
+static INLINE int predict_primary_quant_zero(const tran_low_t *coeff,
```
leolzhao: What about changing the function name to "check_primary_quant_all_zero"?

av2/encoder/tx_search.c
```diff
@@ -2433,6 +2486,14 @@ static void search_tx_type(const AV2_COMP *cpi, MACROBLOCK *x, int plane,
             continue;
           }
         }
+        if (cpi->sf.tx_sf.prune_intra_ist_stx_by_zero_eob && *eob == 0 &&
```
leolzhao: Seems that the code here is quite similar to the code right above. Can you check whether it is possible to combine these two code branches together by adding (cpi->sf.tx_sf.prune_intra_ist_stx_by_zero_eob && *eob == 0) into the above branch?

leolzhao (approved):

av2/encoder/tx_search.c
yeqing-wu: Done

av2/encoder/tx_search.c
yeqing-wu: Combined, the downside is that it introduces an additional check, and the combined code isn’t as straightforward. However, I’m okay with this.

```
        if (plane == PLANE_TYPE_Y && !is_inter &&
            (*eob == 1 ||
             (cpi->sf.tx_sf.prune_intra_ist_stx_by_zero_eob && *eob == 0))) {
          if (tx_type1 == DCT_DCT) eob_found = 1;
          const int kill =
              (*eob == 1) ? (tx_type1 != DCT_DCT || (stx && primary_tx_type))
                          : (stx > 0);
          if (kill) {
            update_txk_array(xd, blk_row, blk_col, tx_size, DCT_DCT);
            continue;
          }
        }
```

PR #5108, opened 2026-07-14
Preserve xlayer contexts across MSDO ID changes

jangsc0000:
## Problem

`stream_info[]` is indexed by the position of an xlayer ID in the current MSDO `stream_ids[]` array.

When the stream ID list changes, the decoder updates `cm->stream_ids` without remapping the existing contexts. This can associate a context with a different xlayer.

For example:

`(0, 2, 4) -> (0, 1, 3) -> (0, 1, 2)`

With the current implementation:

- xlayer 1 can inherit xlayer 2 context.
- xlayer 3 can inherit xlayer 4 context.
- xlayer 2 can later inherit xlayer 3 context.

The [AV2 v1.0 specification](https://github.com/AOMediaCodec/av2-spec/blob/main/v1.0.0/20260528_38f28e7_AV2_Spec_v1.0.0.pdf) allows `sub_xlayer_id` values to change at a random access point without starting a new coded multistream video sequence. Therefore, contexts belonging to xlayer IDs that remain present must not be reset.

Random access points do not need to be aligned across extended layers. For example, during an MSDO transition from `(0, 1, 3)` to `(0, 1, 2)`, xlayer 2 may enter at a random access point while retained xlayers 0 and 1 remain mid-sequence and continue decoding inter frames. Resetting every stream context at this transition would invalidate the DPB references required by xlayers 0 and 1 and break their subsequent inter-frame decoding.

## Changes

- Parse a new MSDO into a temporary configuration before updating `cm->stream_ids`.
- Detect changes to the MSDO stream ID list separately from coded multistream video sequence configuration changes.
- Remap `stream_info[]` by xlayer ID instead of array position.
- Preserve contexts for xlayer IDs that remain present.
- Flush and release contexts for removed xlayer IDs.
- Initialize fresh contexts for newly added xlayer IDs.
- Keep the existing full reset behavior for changes that start a new coded multistream video sequence.

For the example above:

- `(0, 2, 4) -> (0, 1, 3)`: preserve xlayer 0; initialize 1 and 3.
- `(0, 1, 3) -> (0, 1, 2)`: preserve xlayers 0 and 1; initialize 2.
- If a retained xlayer moves to another array position, its context moves with the xlayer ID.

## Tests

Added regression tests covering:

- Context preservation across `(0, 2, 4) -> (0, 1, 3) -> (0, 1, 2)`.
- Context remapping when retained xlayer IDs change array positions.
- Initialization of newly added xlayer IDs.

Validation:

- Debug `test_libavm` build passed.
- `SBEApiTest.*:MsdoStreamContextTest.*`: 19 tests passed.

## Context

This builds on the redundant MSDO handling introduced in #4718 and the configuration-transition flushing added in #4840. Neither change remaps per-stream contexts when only the MSDO `sub_xlayer_id` list changes.

### DIFF
--- a/av2/decoder/obu.c
+++ b/av2/decoder/obu.c
@@ -372,8 +372,10 @@ static void save_msdo_config(const AV2Decoder *pbi, MsdoConfig *config) {
 }
 
 /*!
- * \brief Compare two MSDO configurations
- * \return true if configurations are identical, false otherwise
+ * \brief Compare MSDO fields that define a coded multistream video sequence
+ * boundary. Stream IDs are intentionally excluded because they may change at
+ * a random access point without starting a new coded multistream sequence.
+ * \return true if the sequence-defining fields are identical, false otherwise
  */
 static bool msdo_config_equal(const MsdoConfig *a, const MsdoConfig *b) {
   if (a->multistream_profile_idc != b->multistream_profile_idc) return false;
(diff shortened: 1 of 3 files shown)


peterderivaz:
Related to #5115.  I think this PR is based on a misunderstanding of the design intent, and the code already works correctly.

jangsc0000:
Closing this PR based on the clarification in #5115.

The intended model is that `streamID = i` owns the persistent decoder context, while `sub_xlayer_id[i]` / `obu_xlayer_id` determine the extended layer currently used by that stream. Therefore, remapping `stream_info[]` by numeric xlayer ID, as implemented here, is not the intended behavior.

The discussion in #5115 also records the terminology concern and the limitation of Annex F for changing MSDO mappings.

PR #5111, opened 2026-07-14
Remove FSC path from intra joint UV function

jingninghan:
The intra chroma component will not use FSC mode. Simplify the encoding functions therein.


yeqing-wu (approved):
LGTM

PR #5112, opened 2026-07-14
Add 12-bit profile (Main_444C_12_IP2)

xinzhao-apple:
This PR enables the 12-bit profile by default and brings its implementation into conformance with AV2 Spec Draft with 12-bit support: https://github.com/AOMediaCodec/av2-spec-internal/pull/776.

1. Enable CONFIG_TESTONLY_12BIT_SUPPORT by default
-   Flips the 12-bit config default from 0 to 1 so 12-bit encode/decode is built in by default. (Superseded by the rename below; the flag is now CONFIG_12BIT_PROFILE.)

2. Align MAIN_4xx_12_IP2 profile with AV2 spec Annex A
- Renames CONFIG_TESTONLY_12BIT_SUPPORT to CONFIG_12BIT_PROFILE, and the profile enum TEST_ONLY_12BIT_PROFILE to Main_444C_12_IP2 (spec Main_444C_12_IP2, seq_profile_idc = 5).
- Accepts 8/10/12-bit and all chroma formats (4:0:0/4:2:0/4:2:2/4:4:4) for profile 5, per Table A.1 (previously 8/10-bit were rejected).
- Sets the max embedded-layer count for profile 5 to 3, matching its IOP2 limit (Table A.3).
- Gives profile 5 its own PicSize/Bitrate factor row (36 / 3.0) per Table A.2.
- Updates the in-code Annex A profile/config tables (A.1/A.2/A.5/A.6) and comments, including the bit_depth_idc mapping.

### DIFF
--- a/.github/workflows/build-job-reusable.yaml
+++ b/.github/workflows/build-job-reusable.yaml
@@ -126,7 +126,7 @@ jobs:
               CMAKE_FLAGS="${CMAKE_FLAGS} -DCONFIG_BITSTREAM_DEBUG=1 -DCONFIG_MISMATCH_DEBUG=1"
               ;;
             enable-12bit-profile)
-              CMAKE_FLAGS="${CMAKE_FLAGS} -DCONFIG_TESTONLY_12BIT_SUPPORT=1"
+              CMAKE_FLAGS="${CMAKE_FLAGS} -DCONFIG_12BIT_PROFILE=1"
               ;;
             nasm)
               CMAKE_FLAGS="${CMAKE_FLAGS} -DENABLE_NASM=1"
--- a/.github/workflows/build-job-reusable.yaml
+++ b/.github/workflows/build-job-reusable.yaml
@@ -125,9 +125,6 @@ jobs:
             bitstream-mismatch-debug)
               CMAKE_FLAGS="${CMAKE_FLAGS} -DCONFIG_BITSTREAM_DEBUG=1 -DCONFIG_MISMATCH_DEBUG=1"
               ;;
-            enable-12bit-profile)
-              CMAKE_FLAGS="${CMAKE_FLAGS} -DCONFIG_12BIT_PROFILE=1"
-              ;;
             nasm)
               CMAKE_FLAGS="${CMAKE_FLAGS} -DENABLE_NASM=1"
               ;;
--- a/av2/av2_cx_iface.c
+++ b/av2/av2_cx_iface.c
@@ -867,10 +867,10 @@ static avm_codec_err_t validate_img(avm_codec_alg_priv_t *ctx,
     case AVM_IMG_FMT_I42016: break;
     case AVM_IMG_FMT_I444:
     case AVM_IMG_FMT_I44416:
-      // MAIN_444_10_IP1 and MAIN_4xx_12_IP2 are the profiles that support 444
+      // MAIN_444_10_IP1 and MAIN_444C_12_IP2 are the profiles that support 444
       if (ctx->cfg.g_profile != (unsigned int)MAIN_444_10_IP1 &&
 #if CONFIG_12BIT_PROFILE
-          ctx->cfg.g_profile != (unsigned int)MAIN_4xx_12_IP2 &&
+          ctx->cfg.g_profile != (unsigned int)MAIN_444C_12_IP2 &&
 #endif  // CONFIG_12BIT_PROFILE
           !ctx->cfg.monochrome) {
         ERROR("Invalid image format. I444 images not supported in profile.");
@@ -880,7 +880,7 @@ static avm_codec_err_t validate_img(avm_codec_alg_priv_t *ctx,
     case AVM_IMG_FMT_I42216:
       if (ctx->cfg.g_profile != (unsigned int)MAIN_422_10_IP1
 #if CONFIG_12BIT_PROFILE
-          && ctx->cfg.g_profile != (unsigned int)MAIN_4xx_12_IP2
+          && ctx->cfg.g_profile != (unsigned int)MAIN_444C_12_IP2
 #endif  // CONFIG_12BIT_PROFILE
       ) {
         ERROR("Invalid image format. I422 images not supported in profile.");
--- a/av2/av2_cx_iface.c
+++ b/av2/av2_cx_iface.c
@@ -666,13 +666,13 @@ static avm_codec_err_t validate_config(avm_codec_alg_priv_t *ctx,
   RANGE_CHECK(cfg, g_timebase.num, 1, cfg->g_timebase.den);
 
   RANGE_CHECK_HI(cfg, g_profile, MAX_PROFILES - 1);
-#if !CONFIG_TESTONLY_12BIT_SUPPORT
+#if !CONFIG_12BIT_PROFILE
   RANGE_CHECK(cfg, g_bit_depth, AVM_BITS_8, AVM_BITS_10);
   RANGE_CHECK(cfg, g_input_bit_depth, AVM_BITS_8, AVM_BITS_10);
 #else
   RANGE_CHECK(cfg, g_bit_depth, AVM_BITS_8, AVM_BITS_12);
   RANGE_CHECK(cfg, g_input_bit_depth, AVM_BITS_8, AVM_BITS_12);
-#endif  // !CONFIG_TESTONLY_12BIT_SUPPORT
+#endif  // !CONFIG_12BIT_PROFILE
 
   const int min_quantizer =
       (-(int)(cfg->g_bit_depth - AVM_BITS_8) * MAXQ_OFFSET);
@@ -860,10 +860,6 @@ static avm_codec_err_t validate_config(avm_codec_alg_priv_t *ctx,
 
 static avm_codec_err_t validate_img(avm_codec_alg_priv_t *ctx,
                                     const avm_image_t *img) {
-#if CONFIG_TESTONLY_12BIT_SUPPORT
-  if (ctx->cfg.g_profile == TEST_ONLY_12BIT_PROFILE) return AVM_CODEC_OK;
-#endif  // CONFIG_TESTONLY_12BIT_SUPPORT
-
   switch (img->fmt) {
     case AVM_IMG_FMT_YV12:
     case AVM_IMG_FMT_I420:
@@ -871,15 +867,22 @@ static avm_codec_err_t validate_img(avm_codec_alg_priv_t *ctx,
     case AVM_IMG_FMT_I42016: break;
     case AVM_IMG_FMT_I444:
     case AVM_IMG_FMT_I44416:
-      // MAIN_444_10_IP1 is the only profile that support 444
+      // MAIN_444_10_IP1 and MAIN_4xx_12_IP2 are the profiles that support 444
       if (ctx->cfg.g_profile != (unsigned int)MAIN_444_10_IP1 &&
+#if CONFIG_12BIT_PROFILE
+          ctx->cfg.g_profile != (unsigned int)MAIN_4xx_12_IP2 &&
+#endif  // CONFIG_12BIT_PROFILE
           !ctx->cfg.monochrome) {
         ERROR("Invalid image format. I444 images not supported in profile.");
       }
       break;
     case AVM_IMG_FMT_I422:
     case AVM_IMG_FMT_I42216:
-      if (ctx->cfg.g_profile != (unsigned int)MAIN_422_10_IP1) {
+      if (ctx->cfg.g_profile != (unsigned int)MAIN_422_10_IP1
+#if CONFIG_12BIT_PROFILE
+          && ctx->cfg.g_profile != (unsigned int)MAIN_4xx_12_IP2
+#endif  // CONFIG_12BIT_PROFILE
+      ) {
         ERROR("Invalid image format. I422 images not supported in profile.");
       }
       break;
--- a/av2/common/annexA.c
+++ b/av2/common/annexA.c
@@ -320,15 +320,14 @@ int av2_check_profile_interop_conformance(
 // picture_size_profile_factor_table[] in level.c, and bitrate_profile_factor[]
 // in timing.c).
 //
-// NOTE: This return value is NOT always the spec's ProfileScalingFactor from
-// Table A.2. It coincides with it for profiles 0..4 (0, 0, 0, 1, 2), but
-// profile MAIN_444C_12_IP2 (idc 5) has spec ProfileScalingFactor 2 -- the same
-// as profile MAIN_444_10_IP1 (idc 4) -- while requiring DIFFERENT PicSize and
-// Bitrate factors (36 / 3.0 vs 30 / 2.5). Since a single ProfileScalingFactor
-// value cannot select two different factor rows, profile 5 is given its own
-// row index (3) here. All callers use this value only as a table index, never
-// as the literal ProfileScalingFactor.
-int get_profile_scaling_factor(int seq_profile_idc) {
+// NOTE: This is a table row index, NOT the spec's ProfileScalingFactor from
+// Table A.2. The two coincide for profiles 0..4 (0, 0, 0, 1, 2), but profile
+// MAIN_444C_12_IP2 (idc 5) has spec ProfileScalingFactor 2 -- the same as
+// profile MAIN_444_10_IP1 (idc 4) -- while requiring DIFFERENT PicSize and
+// Bitrate factors (36 / 3.0 vs 30 / 2.5). Since one ProfileScalingFactor value
+// cannot select two different factor rows, profile 5 is given its own row
+// index (3) here.
+int get_profile_factor_table_row_index(int seq_profile_idc) {
   if (seq_profile_idc == MAIN_420_10_IP0 ||
       seq_profile_idc == MAIN_420_10_IP1 ||
       seq_profile_idc == MAIN_420_10_IP2) {
--- a/av2/common/annexA.c
+++ b/av2/common/annexA.c
@@ -27,7 +27,8 @@
  *       0         | C_Main_420_10       | Main    | 8, 10       |  4:0:0, 4:2:0
  *       1         | C_Main_422_10       | Main    | 8, 10       |  4:0:0, 4:2:0, 4:2:2
  *       2         | C_Main_444_10       | Main    | 8, 10       |  4:0:0, 4:2:0, 4:4:4
- *       3-63      | Reserved            | -       | -           |
+ *       3         | C_Main_4xx_12       | Main    | 8, 10, 12   |  4:0:0, 4:2:0, 4:2:2, 4:4:4
+ *       4-63      | Reserved            | -       | -           |
  *
  * Notes:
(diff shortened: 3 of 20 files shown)


av2/common/enums.h
```diff
@@ -271,10 +271,10 @@ enum {
   MAIN_420_10_IP2 = 2,
   MAIN_422_10_IP1 = 3,
   MAIN_444_10_IP1 = 4,
-#if CONFIG_TESTONLY_12BIT_SUPPORT
-  // Fake profile value only for testing 12-bit. Not defined in AV2 spec.
-  TEST_ONLY_12BIT_PROFILE,
-#endif  // CONFIG_TESTONLY_12BIT_SUPPORT
+#if CONFIG_12BIT_PROFILE
+  // 12-bit profile: all chroma formats (4:0:0/4:2:0/4:2:2/4:4:4), IOP2
+  MAIN_4xx_12_IP2,
```
wantehchang: Nit: Assign a value explicitly:

```
  MAIN_4xx_12_IP2 = 5,
```

av2/common/annexA.c
```diff
@@ -49,13 +51,22 @@ typedef enum {
...
+ *                         |                      | CHROMA_FORMAT_420,      |
+ *                         |                      | CHROMA_FORMAT_444       |
+ *   C_Main_4xx_12         | 0..5, 31             | CHROMA_FORMAT_400,      | 0..2
+ *                         |                      | CHROMA_FORMAT_420,      |
+ *                         |                      | CHROMA_FORMAT_422,      |
+ *                         |                      | CHROMA_FORMAT_444       |
  *
  * Notes:
  * - seq_profile_idc: Allowed profile values
- * - bit_depth_idc: 0=8-bit, 1=10-bit
```
wantehchang, line 58: Optional: It would be good to create a separate pull request to correct the comment error. The comment fix could then be cherry-picked to the normative branch.

av2/common/annexA.c
```diff
@@ -198,8 +210,14 @@ int av2_check_profile_interop_conformance(
...
 
-  // All profiles support 8-bit and 10-bit only
+  // All profiles support 8-bit and 10-bit. Profile MAIN_4xx_12_IP2
+  // additionally supports 12-bit.
   int is_valid_bit_depth = check_bit_depth_8_10(bit_depth);
+#if CONFIG_12BIT_PROFILE
+  if (profile == MAIN_4xx_12_IP2 && bit_depth == AVM_BITS_12) {
+    is_valid_bit_depth = 1;
+  }
+#endif  // CONFIG_12BIT_PROFILE
```
wantehchang: Nit: I think it would be better to change `check_bit_depth_8_10()` to take the `profile` parameter as follows:

```
static int check_bit_depth_8_10(int bit_depth, BITSTREAM_PROFILE profile) {
  (void)profile;
  // All profiles support 8-bit and 10-bit. Profile MAIN_4xx_12_IP2
  // additionally supports 12-bit.
  if (bit_depth == AVM_BITS_8 || bit_depth == AVM_BITS_10) {
    return 1;
  }
#if CONFIG_12BIT_PROFILE
  if (profile == MAIN_4xx_12_IP2 && bit_depth == AVM_BITS_12) {
    return 1;
  }
#endif  // CONFIG_12BIT_PROFILE
  return 0;
}
```

Then here we can just say:

```
  int is_valid_bit_depth = check_bit_depth(bit_depth, profile);
```

av2/common/annexA.c
```diff
@@ -252,11 +270,17 @@ int av2_check_profile_interop_conformance(
     case CONFIGURABLE: {
       // Supports all chroma formats
     } break;
+#if CONFIG_12BIT_PROFILE
+    case MAIN_4xx_12_IP2:
+      // Supports all chroma formats (4:0:0/4:2:0/4:2:2/4:4:4) at 8/10/12-bit
+      break;
+#endif  // CONFIG_12BIT_PROFILE
```
wantehchang, line 277: Nit: I suggest adding `case MAIN_4xx_12_IP2` before `case CONFIGURABLE`

av2/av2_cx_iface.c
```diff
@@ -860,9 +860,9 @@ static avm_codec_err_t validate_config(avm_codec_alg_priv_t *ctx,
 
 static avm_codec_err_t validate_img(avm_codec_alg_priv_t *ctx,
                                     const avm_image_t *img) {
-#if CONFIG_TESTONLY_12BIT_SUPPORT
-  if (ctx->cfg.g_profile == TEST_ONLY_12BIT_PROFILE) return AVM_CODEC_OK;
-#endif  // CONFIG_TESTONLY_12BIT_SUPPORT
+#if CONFIG_12BIT_PROFILE
+  if (ctx->cfg.g_profile == MAIN_4xx_12_IP2) return AVM_CODEC_OK;
```
wantehchang: IMPORTANT: This early return is incorrect. We should not skip the last check (at line 893) in this function.

wantehchang (commented):
Xin: Thank you for writing this PR. I suggest some simple changes.

av2/common/enums.h
xinzhao-apple: Fixed. Thanks.

av2/av2_cx_iface.c
xinzhao-apple: Agreed and fixed.

av2/common/annexA.c
xinzhao-apple: Thanks for the suggestion. Agreed this is better, updated this PR.

av2/common/annexA.c
xinzhao-apple: Agreed and updated.

av2/common/annexA.c
xinzhao-apple: @urvangjoshi Any preference on this? I am good either way (include this change in a separate PR or keep it in this PR).

urvangjoshi: Fine either way as well.   Slight preference for a separate PR since this is a separate issue from this PR's intent.

av2/common/annexA.c
```diff
@@ -27,7 +27,8 @@
  *       0         | C_Main_420_10       | Main    | 8, 10       |  4:0:0, 4:2:0
  *       1         | C_Main_422_10       | Main    | 8, 10       |  4:0:0, 4:2:0, 4:2:2
  *       2         | C_Main_444_10       | Main    | 8, 10       |  4:0:0, 4:2:0, 4:4:4
- *       3-63      | Reserved            | -       | -           |
+ *       3         | C_Main_444C_12      | Main    | 8, 10, 12   |  4:0:0, 4:2:0, 4:2:2, 4:4:4
```
wantehchang, line 30: Just curious: what does "444C" mean?

av2/common/annexA.c
```diff
@@ -281,11 +302,26 @@ int av2_check_profile_interop_conformance(
...
+// in timing.c).
+//
+// NOTE: This return value is NOT always the spec's ProfileScalingFactor from
+// Table A.2. It coincides with it for profiles 0..4 (0, 0, 0, 1, 2), but
+// profile MAIN_444C_12_IP2 (idc 5) has spec ProfileScalingFactor 2 -- the same
+// as profile MAIN_444_10_IP1 (idc 4) -- while requiring DIFFERENT PicSize and
+// Bitrate factors (36 / 3.0 vs 30 / 2.5). Since a single ProfileScalingFactor
+// value cannot select two different factor rows, profile 5 is given its own
+// row index (3) here. All callers use this value only as a table index, never
+// as the literal ProfileScalingFactor.
```
wantehchang: Given this comment, should we rename this function `get_profile_factor_table_row_index`?

I.e., remove `_scaling` and add `_row_index`.

av2/common/enums.h
```diff
@@ -271,10 +271,10 @@ enum {
   MAIN_420_10_IP2 = 2,
   MAIN_422_10_IP1 = 3,
   MAIN_444_10_IP1 = 4,
-#if CONFIG_TESTONLY_12BIT_SUPPORT
-  // Fake profile value only for testing 12-bit. Not defined in AV2 spec.
-  TEST_ONLY_12BIT_PROFILE,
-#endif  // CONFIG_TESTONLY_12BIT_SUPPORT
+#if CONFIG_12BIT_PROFILE
+  // 12-bit profile: all chroma formats (4:0:0/4:2:0/4:2:2/4:4:4), IOP2
```
wantehchang: Nit: IOP2 => IP2 ?

av2/common/timing.c
```diff
@@ -52,7 +52,7 @@ static int32_t high_kbps[1 << LEVEL_BITS] = {
 
 /* BitrateProfileFactor */
 static int bitrate_profile_factor[1 << PROFILE_BITS] = {
-  1, 2, 3, 0, 0, 0, 0, 0
+  1, 2, 3, 3, 0, 0, 0, 0
```
wantehchang, line 57: How is this table related to the `bitrate_profile_factor_table` in level.c?

It seems that this table's values are the rounded values of `bitrate_profile_factor_table`?

wantehchang, line 57: Remove the `0` entries in this table. Keep just the first four entries.

av2/common/timing.c
```diff
@@ -52,7 +52,7 @@ static int32_t high_kbps[1 << LEVEL_BITS] = {
 
 /* BitrateProfileFactor */
 static int bitrate_profile_factor[1 << PROFILE_BITS] = {
```
wantehchang, line 56: IMPORTANT: Remove the array size `1 << PROFILE_BITS`. This array is indexed by the row index returned by `get_profile_scaling_factor()`, which is 0..3.

wantehchang (approved):
Xin: LGTM. Thanks! Note that I did not compare this PR with the corresponding spec PR (https://github.com/AOMediaCodec/av2-spec-internal/pull/776),

av2/common/annexA.c
xinzhao-apple: C stands for Comprehensive, this is the naming suggested and agreed during the VCWG meeting.

C_Main_444_12 is not used, because an existing C_Main_444_10 does not support 4:2:0.

av2/common/annexA.c
xinzhao-apple: Ok, I kept this comment unchanged in this PR.

av2/common/annexA.c
xinzhao-apple: Thanks. Updated the function name accordingly.

av2/common/enums.h
xinzhao-apple: Thanks. IOP2 changed to IP2 in the updated PR.

av2/common/timing.c
xinzhao-apple: Done in the updated PR. Thanks.

av2/common/timing.c
xinzhao-apple: Done in the updated PR. Thanks.

av2/common/timing.c
xinzhao-apple: Thanks for pointing it out!

I think both tables implement BitrateProfileFactor from Annex A Table A.2, they're duplicates, and timing.c's copy is stale rather than rounded. timing.c's {1, 2, 3} dates to [2649551c0](https://gitlab.com/AOMediaCodec/avm/-/commit/2649551c0574f0430f0c79ecaed4663c39bee6f1?old_path=av1%2Fencoder%2Ftiming.c&new_path=av1%2Fcommon%2Ftiming.c#line_e637dcfb6_A35) (Jun 2018), when AV1 had three profiles whose factors really were 1/2/3. level.c's table was added later in [45a6216e2](https://gitlab.com/AOMediaCodec/avm/-/commit/45a6216e2c37d019cb9766027ca6c77532ed1ace?file_path=av2%2Fcommon%2Flevel.c#line_7e1f9c86b_A392) (Jan 2026), initially also as {1.0, 2.0, 3.0} consistent at the time. Then [5d2708d1a](https://gitlab.com/AOMediaCodec/avm/-/commit/5d2708d1a7e77c1bc698903b26184aa679dc9642?file_path=av2%2Fcommon%2Flevel.c#line_7e1f9c86b_A486) (CWG-G075/G078, Mar 2026) updated level.c to {1.0, 1.667, 2.5} and PicSize to {15, 20, 30}, but only updated the call site in timing.c, not its value table. That's where they diverged.

So in my understanding, this is a pre-existing on main, unrelated to the 12-bit change in this MR (the entry associated with 12-bit is consistent between two tables), I'd suggest a separate issue and fixing it in a separate PR by the relevant proponent, potentially consolidating these two tables as one single table to avoid divergence. What do you think?

av2/encoder/rd.c
```diff
@@ -700,6 +700,10 @@ int av2_get_deltaq_offset(const AV2_COMP *cpi, int qindex, double beta) {
                            cpi->common.seq_params.bit_depth);
   int newq = (int)rint(q / sqrt(beta));
   int orig_qindex = qindex;
+  const int max_qindex =
+      cpi->common.seq_params.bit_depth == AVM_BITS_8    ? MAXQ_8_BITS
+      : cpi->common.seq_params.bit_depth == AVM_BITS_10 ? MAXQ_10_BITS
+                                                        : MAXQ_12_BITS;
```
wantehchang, line 706: Nice cleanup!

av2/common/timing.c
wantehchang: Xin: Thank you very much for researching this discrepancy. I agree with your suggestion. I will file an issue so that we can deal with this separately.

wantehchang (approved):

av2/common/timing.c
```diff
@@ -50,10 +50,9 @@ static int32_t high_kbps[1 << LEVEL_BITS] = {
   UNDEFINED_RATE, UNDEFINED_RATE, UNDEFINED_RATE, UNDEFINED_RATE
 };
 
-/* BitrateProfileFactor */
-static int bitrate_profile_factor[1 << PROFILE_BITS] = {
-  1, 2, 3, 0, 0, 0, 0, 0
-};
+/* BitrateProfileFactor, indexed by the profile's factor table row
+ * (see get_profile_factor_table_row_index() in annexA.c). */
+static int bitrate_profile_factor[] = { 1, 2, 3, 3 };
```
wantehchang: Xin: While filing issue https://github.com/AOMediaCodec/avm/issues/5179 on this table, I found two things:
- This table should be indexed by `seq_profile_idc` directly, because the decoder uses this table to detect unsupported profiles.
- The 0 entries in this table are important. They indicate an unsupported profile.

Could you please change the table dimension back to `1 << PROFILE_BITS` and add back the 0 entries?

I will ask you to review a pull request that changes `av2_max_level_bitrate()` to index this table with `seq_profile` directly.

av2/common/timing.c
xinzhao-apple: Ok, sounds good to me.

av2/common/timing.c
wantehchang: Xin: Please disregard my previous comment completely.

While writing a fix for this issue, I found a better solution that does not require changing the code here. Please see PR https://github.com/AOMediaCodec/avm/pull/5181.

av2/common/timing.c
wantehchang: I filed issue https://github.com/AOMediaCodec/avm/issues/5179.

av2/encoder/rd.c
xinzhao-apple: Thanks.

av2/common/timing.c
```diff
@@ -50,9 +50,11 @@ static int32_t high_kbps[1 << LEVEL_BITS] = {
   UNDEFINED_RATE, UNDEFINED_RATE, UNDEFINED_RATE, UNDEFINED_RATE
 };
 
-/* BitrateProfileFactor */
+/* BitrateProfileFactor. A zero entry indicates an unsupported profile: callers
+ * treat a zero bitrate as "profile, level, and tier combination not
+ * supported". */
```
wantehchang: I suggest we restore your previous comment partially and move the last part of the new comment to the `av2_max_level_bitrate()` function, as follows:

```
/* BitrateProfileFactor, indexed by the profile's factor table row
 * (see get_profile_factor_table_row_index() in annexA.c).
 * A zero entry indicates an unsupported profile. */
static int bitrate_profile_factor[1 << PROFILE_BITS] = {
  1, 2, 3, 3, 0, 0, 0, 0
};

// Callers treat a zero bitrate as "profile, level, and tier combination not supported".
int64_t av2_max_level_bitrate(BITSTREAM_PROFILE seq_profile, int seq_level_idx,
```

wantehchang (approved):

av2/common/timing.c
xinzhao-apple: Updated this part.

wantehchang (approved):

.github/workflows/build-job-reusable.yaml
```diff
@@ -126,7 +126,7 @@ jobs:
               CMAKE_FLAGS="${CMAKE_FLAGS} -DCONFIG_BITSTREAM_DEBUG=1 -DCONFIG_MISMATCH_DEBUG=1"
               ;;
             enable-12bit-profile)
-              CMAKE_FLAGS="${CMAKE_FLAGS} -DCONFIG_TESTONLY_12BIT_SUPPORT=1"
+              CMAKE_FLAGS="${CMAKE_FLAGS} -DCONFIG_12BIT_PROFILE=1"
```
urvangjoshi: Now that this flag is on by default,  this CI check is redundant. 

If desired, the CI check should be modified to be `disable-12bit-profile`  instead which sets `-DCONFIG_12BIT_PROFILE=0`

urvangjoshi (approved):
One comment related to CI job -- otherwise looks good.

.github/workflows/build-job-reusable.yaml
wantehchang: If the `CONFIG_12BIT_PROFILE` flag will be removed when the next version of AV2 spec is published, we can simply delete this CI check.

urvangjoshi: That will be fine.  Pls go ahead and remove this CI job by searching for `enable-12bit-profile` in the codebase: https://github.com/search?q=repo%3AAOMediaCodec%2Favm+enable-12bit-profile&type=code

xinzhao-apple: I have removed the redundant CI check in the updated PR.

wantehchang (approved):

urvangjoshi:
@xinzhao-apple @wantehchang :  the PR contains 9 commits.  When merging,  do we want Github to squash them into a single commit?  Or keep 9 separate commits?

wantehchang:
Urvang: Please wait for Xin's reply. I think the 9 commits shoud be squashed into a single commit when you merge this PR.

xinzhao-apple:
> @xinzhao-apple @wantehchang : the PR contains 9 commits. When merging, do we want Github to squash them into a single commit? Or keep 9 separate commits?

Hi @urvangjoshi I think we should squash them into a single commit. Thanks.

PR #5113, opened 2026-07-14
Fix the cast for the BITSTREAM_PROFILE profile var

wantehchang:
In the av2_check_profile_interop_conformance() function, the local variable `profile` is of the BITSTREAM_PROFILE type, so it is redundant to cast `profile` to BITSTREAM_PROFILE. I suspect the intended cast is `int`, to match the %d printf format specifier.

Alternatively, we can simply remove the BITSTREAM_PROFILE casts.

### DIFF
--- a/av2/common/annexA.c
+++ b/av2/common/annexA.c
@@ -218,7 +218,7 @@ int av2_check_profile_interop_conformance(
             is_decoder ? AVM_CODEC_UNSUP_BITSTREAM : AVM_CODEC_INVALID_PARAM,
             "Profile %d only supports YUV 4:0:0 and YUV 4:2:0 color formats, "
             "but color format YUV %s was provided.",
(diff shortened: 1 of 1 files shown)


av2/common/annexA.c
```diff
@@ -218,7 +218,7 @@ int av2_check_profile_interop_conformance(
             is_decoder ? AVM_CODEC_UNSUP_BITSTREAM : AVM_CODEC_INVALID_PARAM,
             "Profile %d only supports YUV 4:0:0 and YUV 4:2:0 color formats, "
             "but color format YUV %s was provided.",
-            (BITSTREAM_PROFILE)profile, is_422 ? "4:2:2" : "4:4:4");
+            (int)profile, is_422 ? "4:2:2" : "4:4:4");
```
wantehchang, line 221: I noticed this while reviewing Xin's PR https://github.com/AOMediaCodec/avm/pull/5112.

urvangjoshi (approved):

PR #5114, opened 2026-07-14
Fix latent assert in handle_newmv MV-precision cache lookup

yeqing-wu:
Tighten the loop lower bound to pb_mv_precision + 1 so the found cache precision is strictly greater than the current pb_mv_precision, matching the assert's intent.

### DIFF
--- a/av2/encoder/rdopt.c
+++ b/av2/encoder/rdopt.c
@@ -1388,7 +1388,7 @@ static int64_t handle_newmv(const AV2_COMP *const cpi, MACROBLOCK *const x,
                        is_pb_mv_precision_active(&cpi->common, mbmi, bsize);
     if (do_refine_ms) {
       int valid_mv0_found = 0;
-      for (int prev_mv_precision = pb_mv_precision;
+      for (int prev_mv_precision = pb_mv_precision + 1;
            prev_mv_precision <= mbmi->max_mv_precision; prev_mv_precision++) {
         assert(get_ref_mv_idx(mbmi, 1) == get_ref_mv_idx(mbmi, 0));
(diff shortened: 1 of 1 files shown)


yeqing-wu:
The main branch also faces the same issue. I’m not sure if we should merge this PR into the main branch as well.

urvangjoshi:
Does this result in a stat change @yeqing-wu ? Curious as we are modifying the for loop.

yeqing-wu:
> Does this result in a stat change @yeqing-wu ? Curious as we are modifying the for loop.

I didn’t include STATS_CHANGED in the commit message, and it passed the pipeline test, so I think it didn’t cause the BD rate change. I didn’t run the CTC tests for this because it’s a potential bug. @urvangjoshi

urvangjoshi (approved):

urvangjoshi:
> The main branch also faces the same issue. I’m not sure if we should merge this PR into the main branch as well.

I think it will be good to cherry-pick this to main branch.

In general, perhaps we need to restrict such cherry-picks between these 2 branches,  so eventual merge of av2-enc branch onto main has less conflicts.   But these kind of small PRs should be OK to cherry-pick right away I think.

yeqing-wu:
> > The main branch also faces the same issue. I’m not sure if we should merge this PR into the main branch as well.
> 
> I think it will be good to cherry-pick this to main branch.
> 
> In general, perhaps we need to restrict such cherry-picks between these 2 branches, so eventual merge of av2-enc branch onto main has less conflicts. But these kind of small PRs should be OK to cherry-pick right away I think.

Thanks @urvangjoshi, let’s keep it in av2-enc branch until we reach the milestone and merge all into the main branch. This false alarm assertion occurred occasionally; I encountered it twice, so it should be a rare case, and we can leave it in av2-enc right now.

PR #5116, opened 2026-07-15
Fix setting in multi_layers tests

marpan-max:
For the MultiLayerTest2Embedded2TemporaSframe test, which has dynamic layer dropping,
the flag enable_explicit_ref_frame_map_ should be set to true, to avoid DPB desync.

Failures were seen at the lower speed settings (e.g., speed 2), not at the speed 5 setting used in the test.

### DIFF
--- a/test/multi_layer_test.cc
+++ b/test/multi_layer_test.cc
@@ -464,17 +464,6 @@ TEST_P(MultiLayerTest, MultiLayerTest2Embedded) {
   EXPECT_EQ(num_mismatch_, 0);
 }
 
-TEST_P(MultiLayerTest, MultiLayerTest2EmbeddedDecodeBaseOnlyImplicit) {
-  ::libavm_test::Y4mVideoSource video_nonsc("park_joy_90p_8_420.y4m", 0, 20);
-  num_temporal_layers_ = 1;
-  num_embedded_layers_ = 2;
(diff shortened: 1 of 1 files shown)


yeqing-wu:
Hi @marpan-max,

Thanks for cherry-picking this fix from the main branch to the av2-enc branch. I’ve also noticed that there are failures in the multi-layer unit tests, which occur quite frequently. Apart from this particular issue, do you think that other unit tests in test/multi_layer_test.cc could be causing the same failure? Since there are several tests that also set `enable_explicit_ref_frame_map_ = false`? Any insight?

Thanks
Yeqing

marpan-max:
Updated the patch, with these changes I don't see any more issues with these tests regarding the setting enable_explicit_ref_frame_map_

yeqing-wu:
Thanks for updating the fix. The two changes you made (`MultiLayerTest2EmbeddedDecodeBaseOnlyImplicit` and `MultiLayerTest3EmbeddedDecodeBaseOnlyImplicit`) were the ones that I noticed frequently causing the failure..

yeqing-wu:
Since this code is cherry-picked from the main branch, which has been reviewed . Thus merging to av2-enc to resolve unit test failures.

yeqing-wu (approved):

PR #5118, opened 2026-07-16
Rework av2_encode_block_intra_joint_uv

jingninghan:
Make it properly handle cctx cases.

Take out coeff dropout to be consistent with what's used in RD search stage.

Re-do u plane if its eob is different from that's in RD search to avoid potential mismatch issues due to the change of cctx_type.

Tested for A2 and A4 in speed 1 with 33 frames. No coding stats change observed.

### DIFF
--- a/av2/encoder/encodemb.c
+++ b/av2/encoder/encodemb.c
@@ -1708,67 +1708,36 @@ void av2_encode_block_intra_joint_uv(int block, int blk_row, int blk_col,
   av2_setup_quant(tx_size, use_trellis, quant_idx,
                   cpi->oxcf.q_cfg.quant_b_adapt, &quant_param);
 
-  // Settings for optimization type. NOTE: To set optimization type for all
-  // intra frames, both `KEY_BLOCK_OPT_TYPE` and `INTRA_BLOCK_OPT_TYPE` should
-  // be set.
-  // TODO(yjshen): These settings are hard-coded and look okay for now. They
-  // should be made configurable later.
-  // Blocks of key frames ONLY.
-  OPT_TYPE KEY_BLOCK_OPT_TYPE = TRELLIS_DROPOUT_OPT;
-  // Blocks of intra frames (key frames EXCLUSIVE).
-  OPT_TYPE INTRA_BLOCK_OPT_TYPE = TRELLIS_DROPOUT_OPT;
-
-  int use_tcq = cm->features.tcq_mode != 0;
-  if (use_tcq) {
-    // Dropout setting should be disabled when Trellis Coded Quant is
-    // enabled.
-    KEY_BLOCK_OPT_TYPE = TRELLIS_OPT;
-    // Blocks of intra frames (key frames EXCLUSIVE).
-    INTRA_BLOCK_OPT_TYPE = TRELLIS_OPT;
-  }
-
-  // Whether trellis or dropout optimization is required for key frames and
-  // intra frames.
-  const bool do_trellis = (frame_is_intra_only(cm) &&
-                           (KEY_BLOCK_OPT_TYPE == TRELLIS_OPT ||
-                            KEY_BLOCK_OPT_TYPE == TRELLIS_DROPOUT_OPT)) ||
-                          (!frame_is_intra_only(cm) &&
-                           (INTRA_BLOCK_OPT_TYPE == TRELLIS_OPT ||
-                            INTRA_BLOCK_OPT_TYPE == TRELLIS_DROPOUT_OPT));
-  const bool do_dropout = (frame_is_intra_only(cm) &&
-                           (KEY_BLOCK_OPT_TYPE == DROPOUT_OPT ||
-                            KEY_BLOCK_OPT_TYPE == TRELLIS_DROPOUT_OPT)) ||
-                          (!frame_is_intra_only(cm) &&
-                           (INTRA_BLOCK_OPT_TYPE == DROPOUT_OPT ||
-                            INTRA_BLOCK_OPT_TYPE == TRELLIS_DROPOUT_OPT));
   for (int plane = AVM_PLANE_U; plane <= AVM_PLANE_V; plane++) {
-    if (plane == AVM_PLANE_V && !is_inter_block(xd->mi[0], xd->tree_type) &&
-        *eob_c1 == 1) {
-      update_cctx_array(xd, blk_row, blk_col, 0, 0,
-                        args->dry_run ? TX_4X4 : tx_size, CCTX_NONE);
-      cctx_type = av2_get_cctx_type(xd, blk_row, blk_col);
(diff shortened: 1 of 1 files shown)


urvangjoshi:
Will wait for your review @kslu-aom .

av2/encoder/encodemb.c
```diff
@@ -1708,67 +1708,37 @@ void av2_encode_block_intra_joint_uv(int block, int blk_row, int blk_col,
...
-      // In dry run, cctx type will not be referenced by neighboring blocks,
-      // so there is no need to fill in the whole chroma region. In addition,
-      // ctx->cctx_type_map size in dry run may not be aligned with actual
-      // chroma coding region for some partition types.
+    if (plane == AVM_PLANE_V && (*eob_c1 == 0 || *eob_c1 == 1) &&
+        cctx_type != CCTX_NONE) {
+      // If the encoding stage uses a different cctx_type from the one
+      // used in the rd stage, re-do the u plane with the updated cctx_type.
       update_cctx_array(xd, blk_row, blk_col, 0, 0,
                         args->dry_run ? TX_4X4 : tx_size, CCTX_NONE);
```
kslu-aom: This CCTX_NONE update was a hack I added to fix a cctx context being messed up. Have you checked if this update is still necessary? If not, we can remove it to avoid any confusion, but we can also do it later if you have not checked.

kslu-aom (approved):

av2/encoder/encodemb.c
jingninghan: It seems still necessary in the event that we need to alter the cctx_type from what the buffer previously has.

yeqing-wu: The code doesn’t refresh `cctx_type` after updating the `cctx_array`. Is this necessary since in the old code, `cctx_type` was refreshed after updating the `cctx_array`?

av2/encoder/encodemb.c
```diff
@@ -1708,67 +1708,37 @@ void av2_encode_block_intra_joint_uv(int block, int blk_row, int blk_col,
...
+      if (quant_param.use_optimize_b) {
+        const ENTROPY_CONTEXT *a = &args->ta[blk_col];
+        const ENTROPY_CONTEXT *l = &args->tl[blk_row];
+        TXB_CTX txb_ctx;
+        get_txb_ctx(plane_bsize, tx_size, AVM_PLANE_U, a, l, &txb_ctx,
+                    xd->mi[0]->fsc_mode[xd->tree_type == CHROMA_PART] &&
+                        cm->seq_params.enable_fsc);
+        av2_optimize_b(args->cpi, x, AVM_PLANE_U, block, tx_size, tx_type,
+                       cctx_type, &txb_ctx, &dummy_rate_cost);
+      }
```
yeqing-wu, line 1732: In the original code, only if the plane is `AVM_PLANE_V` and the value of `*eob_c1` is 1, would the code of `av2_setup_qmatrix(), av2_xform_quant()`, and all the code under the `if (quant_param.use_optimize_b)` condition be called again. However, in the new code, even if `*eob_c1` is 0, the above code would still be called. Is this deviation from the original code correct?

av2/encoder/encodemb.c
```diff
@@ -1708,67 +1708,37 @@ void av2_encode_block_intra_joint_uv(int block, int blk_row, int blk_col,
...
                         args->dry_run ? TX_4X4 : tx_size, CCTX_NONE);
+      av2_setup_xform(cm, x, AVM_PLANE_U, tx_size, tx_type, cctx_type,
+                      &txfm_param);
+      av2_setup_qmatrix(&cm->quant_params, xd, AVM_PLANE_U, tx_size, tx_type,
+                        &quant_param);
+      av2_xform_quant(cm, x, AVM_PLANE_U, block, blk_row, blk_col, plane_bsize,
+                      &txfm_param, &quant_param);
+      if (quant_param.use_optimize_b) {
+        const ENTROPY_CONTEXT *a = &args->ta[blk_col];
+        const ENTROPY_CONTEXT *l = &args->tl[blk_row];
```
yeqing-wu, line 1725: In the original code, there was a plane offset for both locations. However, we have removed the plane offset in this place, and these code are for AVM_PLANE_V plane. Is this correct?

```
        const ENTROPY_CONTEXT *a =
            &args->ta[blk_col + (plane - AVM_PLANE_U) * MAX_MIB_SIZE];
        const ENTROPY_CONTEXT *l =
            &args->tl[blk_row + (plane - AVM_PLANE_U) * MAX_MIB_SIZE];
```

av2/encoder/encodemb.c
kslu-aom: I have a similar question. The issue mentioned in my previous comment does not exist anymore (neighboring cctx_type is not used as a context now). I suspect this update is not necessary now, but it may be ok to leave it as long as there is no stats difference. I will double check later.

yeqing-wu: I believe it’s better to figure out the issue since the design is quite different.
1) In the original code, if these conditions are met, we set cctx_type to CCTX_NONE, and this CCTX_NONE is used in subsequent code.
2) With the new code, we don’t set cctx_type to CCTX_NONE, so it will use the existing cctx_type for subsequent code.

If the code changes don’t cause a change in BD rate, does it mean that the conditions in this code are no longer working? Or is it because we haven’t run enough tests, so these conditions are never triggered?

jingninghan: cctx_type needs to be updated to CCTX_NONE. I updated it in the new version.

Since the cctx_type is not needed as context for later use, the array update can be removed.
I will take it out in a next CL.

jingninghan: The current RD process would force CCTX_NONE if eob_1 is 0 or 1. So this re-run part won't be triggered.

It might not be the case when more speed features are applied around transform search.

av2/encoder/encodemb.c
jingninghan: plane is AVM_PLANE_U, as it re-runs the U plane processing to apply CCTX_NONE, if cctx_type is foced to change from its initial run.

av2/encoder/encodemb.c
jingninghan: This logic is different from the original.

Previously if eob1 is 0, it will force eob2 to be 0 too.

However, the spec allows it to have the following condition:
cctx_type = CCTX_NONE;
eob1 = 0;
eob2 != 0;

The new logic allows the encoder to run quantization for V plane, even if eob1 = 0.

yeqing-wu: Thanks for the explanation.

av2/encoder/encodemb.c
yeqing-wu: I see, I did not notice it is for `AVM_PLANE_U`. Sorry for this.

yeqing-wu (approved):

leolzhao (approved):
LGTM.

PR #5119, opened 2026-07-16
Remove mlayer/implicit-ref-map TIP suppression

yeqing-wu:
The unit tests that exercised base-only implicit-ref-map decode of a multi-layer bitstream (the only path where a TIP-winning base-layer block would fail to reconstruct) were removed in 9f0dfe5 (Fix setting in multi_layers tests, #5116), so this encoder-side workaround is no longer needed.

### DIFF
--- a/av2/encoder/rdopt.c
+++ b/av2/encoder/rdopt.c
@@ -9008,17 +9008,7 @@ void av2_rd_pick_inter_mode_sb(struct AV2_COMP *cpi,
       cm->current_frame.reference_mode == REFERENCE_MODE_SELECT;
 
   const int num_total_refs = cm->ref_frames_info.num_total_refs;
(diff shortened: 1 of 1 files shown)


yeqing-wu:
Hi @leolzhao,

The base layer constraints of MLayer coding are no longer necessary for TIP (the ones you weren’t quite understanding) after the base-only implicit-reference-map decode of a multi-layer bitstream was removed in https://github.com/AOMediaCodec/avm/commit/9f0dfe531e9519420157beb71bd85ca7e498b9c0 (Fix setting in multi_layers tests, https://github.com/AOMediaCodec/avm/pull/5116). Therefore, this encoder-side workaround is no longer required.

I’ve also removed it to aid in understanding. Could you please review it?

Thanks
Yeqing

leolzhao (approved):
LGTM.

PR #5125, opened 2026-07-20
Hybrid search of multi way transform in rough search and winner mode search

jadhuran:
Enable winner transform search

Add small fix for winner transform

Disable 4 way and 5 way transform in rough search and enable in winner

Set winner transform candidates to three

Disable existing winner transform speed features.

Tests - cpu-used =1
Anchor: 6e71b0c85924e61d9403b6222921cd4bb8cf870e
Results (RA): 

A1 - 17 frames
A2 - 33 frames

<pre>
| A1      | 0.13% | 0.01%  | 0.20% | 0.13% | 93.6%    | 100%   |
| A2      | 0.17% | -0.15% | 0.06% | 0.15% | 92.3%    | 100%   |
</pre>

### DIFF
--- a/av2/encoder/rdopt.c
+++ b/av2/encoder/rdopt.c
@@ -6851,7 +6851,8 @@ static AVM_INLINE MB_MODE_INFO *get_winner_mode_stats(
     MACROBLOCK *x, MB_MODE_INFO *best_mbmode, RD_STATS *best_rd_cost,
     int best_rate_y, int best_rate_uv, RD_STATS **winner_rd_cost,
     int *winner_rate_y, int *winner_rate_uv, PREDICTION_MODE *winner_mode,
-    MULTI_WINNER_MODE_TYPE multi_winner_mode_type, int mode_idx) {
+    MULTI_WINNER_MODE_TYPE multi_winner_mode_type, int mode_idx,
+    const AV2_COMMON *const cm, BLOCK_SIZE bsize) {
   MB_MODE_INFO *winner_mbmi;
   if (multi_winner_mode_type) {
     assert(mode_idx >= 0 && mode_idx < x->winner_mode_count);
@@ -6862,12 +6863,30 @@ static AVM_INLINE MB_MODE_INFO *get_winner_mode_stats(
     *winner_rate_y = winner_mode_stat->rate_y;
     *winner_rate_uv = winner_mode_stat->rate_uv;
     *winner_mode = winner_mode_stat->mode;
+    MACROBLOCKD *xd = &x->e_mbd;
+    if (is_warp_mode(winner_mbmi->motion_mode)) {
+      if (!winner_mbmi->wm_params[0].invalid)
+        assign_warpmv(cm, xd->submi, bsize, &winner_mbmi->wm_params[0],
+                      xd->mi_row, xd->mi_col, 0);
+      if (!winner_mbmi->wm_params[1].invalid)
+        assign_warpmv(cm, xd->submi, bsize, &winner_mbmi->wm_params[1],
+                      xd->mi_row, xd->mi_col, 1);
+    }
   } else {
     winner_mbmi = best_mbmode;
     *winner_rd_cost = best_rd_cost;
     *winner_rate_y = best_rate_y;
     *winner_rate_uv = best_rate_uv;
     *winner_mode = best_mbmode->mode;
+    MACROBLOCKD *xd = &x->e_mbd;
+    if (is_warp_mode(winner_mbmi->motion_mode)) {
+      if (!winner_mbmi->wm_params[0].invalid)
+        assign_warpmv(cm, xd->submi, bsize, &winner_mbmi->wm_params[0],
+                      xd->mi_row, xd->mi_col, 0);
+      if (!winner_mbmi->wm_params[1].invalid)
+        assign_warpmv(cm, xd->submi, bsize, &winner_mbmi->wm_params[1],
+                      xd->mi_row, xd->mi_col, 1);
+    }
   }
   return winner_mbmi;
 }
@@ -6910,7 +6929,7 @@ static AVM_INLINE void refine_winner_mode_tx(
     MB_MODE_INFO *winner_mbmi = get_winner_mode_stats(
         x, best_mbmode, rd_cost, best_rate_y, best_rate_uv, &winner_rd_stats,
         &winner_rate_y, &winner_rate_uv, &winner_mode,
-        cpi->sf.winner_mode_sf.multi_winner_mode_type, mode_idx);
+        cpi->sf.winner_mode_sf.multi_winner_mode_type, mode_idx, cm, bsize);
 
     if (xd->lossless[winner_mbmi->segment_id] == 0 &&
         winner_mode != MODE_INVALID &&
@@ -6922,6 +6941,11 @@ static AVM_INLINE void refine_winner_mode_tx(
       const int skip_ctx = av2_get_skip_txfm_context(xd);
 
       *mbmi = *winner_mbmi;
+      struct macroblockd_plane *p = xd->plane;
+      const BUFFER_SET orig_dst = {
+        { p[0].dst.buf, p[1].dst.buf, p[2].dst.buf },
+        { p[0].dst.stride, p[1].dst.stride, p[2].dst.stride },
+      };
 
       set_ref_ptrs(cm, xd, mbmi->ref_frame[0], mbmi->ref_frame[1]);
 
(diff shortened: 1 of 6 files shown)


av2/encoder/rdopt.c
```diff
@@ -9256,12 +9280,18 @@ void av2_rd_pick_inter_mode_sb(struct AV2_COMP *cpi,
 
   int winner_mode_count =
       cpi->sf.winner_mode_sf.multi_winner_mode_type ? x->winner_mode_count : 1;
-  // In effect only when fast tx search speed features are enabled.
-  refine_winner_mode_tx(cpi, x, rd_cost, bsize, ctx, &search_state.best_mbmode,
-                        yv12_mb, search_state.best_rate_y,
-                        search_state.best_rate_uv, &search_state.best_skip2,
-                        winner_mode_count);
 
+  if (!(winner_mode_count == 1 && search_state.best_mode_skippable)) {
```
yunqingwang1, line 9289: Could you explain why "if (!(winner_mode_count == 1 && search_state.best_mode_skippable))" is needed? Thanks.

jadhuran: When winner mode count is 1, but the best mode is not skip, we still have to perform winner transform search for that one mode that was selected. If it is skip, it is of no use to do further search.

av2/encoder/tx_search.c
```diff
@@ -3074,6 +3074,22 @@ static INLINE bool prune_tx_part_eval_using_none_rd(
...
+    TX_PARTITION_TYPE type, int eval_mode_type, REGION_TYPE region_type,
+    int disable_multiway_tx_part_in_rough_mode) {
+  if (!disable_multiway_tx_part_in_rough_mode) return 0;
+
+  if (eval_mode_type == MODE_EVAL && region_type == MIXED_INTER_INTRA_REGION &&
+      type > TX_PARTITION_VERT)
+    return 1;
+
+  if (eval_mode_type == WINNER_MODE_EVAL &&
+      region_type == MIXED_INTER_INTRA_REGION && type <= TX_PARTITION_VERT)
```
yunqingwang1, line 3087: Skipping "type <= TX_PARTITION_VERT" since they are already evaluated during MODE_EVAL pass, right?

jadhuran: yes, we have already evaluated other tx partition types in mode eval.

yunqingwang1:
Nice speedup! Looks like this feature is for speed >= 1. Question for all: do we require to run tests for speed > 1 as well?

av2/encoder/rdopt.c
```diff
@@ -6856,12 +6857,30 @@ static AVM_INLINE MB_MODE_INFO *get_winner_mode_stats(
     *winner_rate_y = winner_mode_stat->rate_y;
     *winner_rate_uv = winner_mode_stat->rate_uv;
     *winner_mode = winner_mode_stat->mode;
+    MACROBLOCKD *xd = &x->e_mbd;
+    if (is_warp_mode(winner_mbmi->motion_mode)) {
+      if (!winner_mbmi->wm_params[0].invalid)
+        assign_warpmv(cm, xd->submi, bsize, &winner_mbmi->wm_params[0], mi_row,
+                      mi_col, 0);
+      if (!winner_mbmi->wm_params[1].invalid)
```
yeqing-wu, line 6871: Do we need to check if it’s compound warp mode? If not, then we shouldn’t use wm_params[1], right?

av2/encoder/rdopt.c
```diff
@@ -6856,12 +6857,30 @@ static AVM_INLINE MB_MODE_INFO *get_winner_mode_stats(
...
     *winner_rd_cost = best_rd_cost;
     *winner_rate_y = best_rate_y;
     *winner_rate_uv = best_rate_uv;
     *winner_mode = best_mbmode->mode;
+    MACROBLOCKD *xd = &x->e_mbd;
+    if (is_warp_mode(winner_mbmi->motion_mode)) {
+      if (!winner_mbmi->wm_params[0].invalid)
+        assign_warpmv(cm, xd->submi, bsize, &winner_mbmi->wm_params[0], mi_row,
+                      mi_col, 0);
+      if (!winner_mbmi->wm_params[1].invalid)
```
yeqing-wu, line 6886: Same comment here

av2/encoder/rdopt.c
```diff
@@ -6904,7 +6923,8 @@ static AVM_INLINE void refine_winner_mode_tx(
     MB_MODE_INFO *winner_mbmi = get_winner_mode_stats(
         x, best_mbmode, rd_cost, best_rate_y, best_rate_uv, &winner_rd_stats,
         &winner_rate_y, &winner_rate_uv, &winner_mode,
-        cpi->sf.winner_mode_sf.multi_winner_mode_type, mode_idx);
+        cpi->sf.winner_mode_sf.multi_winner_mode_type, mode_idx, cm, bsize,
+        xd->mi_row, xd->mi_col);
```
yeqing-wu: Since x is passed as an input parameter, and xd is an element of x, should we retrieve mi_row, mi_col, and size within the get_winner_mode_stats() function to prevent the function from having an excessive number of parameters, thereby enhancing its readability?

yeqing-wu (commented):
Great optimization! I have only a few small comments.

av2/encoder/rdopt.c
jadhuran: Since this is a place to copy modes, we do not need to check mode type. If the mode is not compound type, use wm_params[1] will not be used.

av2/encoder/rdopt.c
jadhuran: Same reply as above.

av2/encoder/rdopt.c
jadhuran: Ok, I will update it.

av2/encoder/rdopt.c
jadhuran: If needed, I can add that condition as well

av2/encoder/rdopt.c
jadhuran: Addressed

av2/encoder/rdopt.c
yeqing-wu: Sure, we can keep it as it is. I understand what you meant. 

By the way, in the AV2 specification, the editors mentioned that “invalid” is not required, so it was entirely removed from the specification. They also suggested removing “invalid” from the reference software. I’m not sure if we need to revisit this in the future.

yeqing-wu (approved):

av2/encoder/rdopt.c
jadhuran: I am also not sure about this, If necessary, we can revisit in the future.

yeqing-wu:
@leolzhao, @jadhuran, @yunqingwang1 

A general question: if attaching an image to the GitHub message, it won’t be displayed correctly after merging. Should we avoid attaching images to the GitHub message since they’ll be used as the final commit message? I think it’s better to use ASCII tables in the Git commit message. You can add ``` before and after the result table to prevent GitHub from messing up the results.

image

Thanks
Yeqing

leolzhao:
> @leolzhao, @jadhuran, @yunqingwang1
> 
> A general question: if attaching an image to the GitHub message, it won’t be displayed correctly after merging. Should we avoid attaching images to the GitHub message since they’ll be used as the final commit message? I think it’s better to use ASCII tables in the Git commit message. You can add ``` before and after the result table to prevent GitHub from messing up the results.
> 
> image
> Thanks Yeqing

Agreed, let's use ASCII tables or texts to describe the performance. @jadhuran

jadhuran:
> > @leolzhao, @jadhuran, @yunqingwang1
> > A general question: if attaching an image to the GitHub message, it won’t be displayed correctly after merging. Should we avoid attaching images to the GitHub message since they’ll be used as the final commit message? I think it’s better to use ASCII tables in the Git commit message. You can add ``` before and after the result table to prevent GitHub from messing up the results.
> > image
> > Thanks Yeqing
> 
> Agreed, let's use ASCII tables or texts to describe the performance. @jadhuran

Addressed here.

yunqingwang1:
@jadhuran  @leolzhao @yeqing-wu Regarding my previous question, I don't have a strong opinion about requesting speed > 1 result. We could only request speed = 1 result if you all agree. The higher speed performance can be tuned later easily anyway.

jadhuran:
> @jadhuran @leolzhao @yeqing-wu Regarding my previous question, I don't have a strong opinion about requesting speed > 1 result. We could only request speed = 1 result if you all agree. The higher speed performance can be tuned later easily anyway.

Thanks, I agree.

jadhuran:
> @leolzhao, @jadhuran, @yunqingwang1
> 
> A general question: if attaching an image to the GitHub message, it won’t be displayed correctly after merging. Should we avoid attaching images to the GitHub message since they’ll be used as the final commit message? I think it’s better to use ASCII tables in the Git commit message. You can add ``` before and after the result table to prevent GitHub from messing up the results.
> 
> image
> Thanks Yeqing

Also, I think there should be way to avoid the results in the commit message when merging. With results, the commit message may be too long. However, we can have results in ASCII format.

yunqingwang1 (approved):

PR #5126, opened 2026-07-20
Skip temporary prediction for opfl modes.

jadhuran:
Skip process_compound_inter() function for opfl modes.

Tests - cpu-used =1
Results (RA):

A1 - 17 frames
A2 - 33 frames

MR7

### DIFF
--- a/av2/encoder/rdopt.c
+++ b/av2/encoder/rdopt.c
@@ -4887,10 +4887,15 @@ static void evaluate_inter_predictor(AV2_COMP *const cpi,
   }
 
   int compmode_interinter_cost = 0;
-
+  int is_opfl_mode =
+      (cpi->sf.inter_sf.skip_temporary_pred_for_opfl &&
+       mbmi->mode >= NEAR_NEARMV_OPTFLOW && mbmi->mode < COMP_INTER_MODE_END)
+          ? 1
+          : 0;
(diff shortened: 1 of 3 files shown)


yeqing-wu:
Hi @jadhuran,

Thanks for your optimizations. I’m curious about the OPFL mode. I believe we always use the default COMPOUND_TYPE (COMPOUND_AVERAGE) and the default interpolation filter (MULTITAP_SHARP) for OPFL modes, so we don’t need to search for the compound weight mode and interpolation filter. Why does this code change save nearly 2% encoding time? Is it because the current code still conducts unnecessary searches?

Thanks
Yeqing

jadhuran:
> Yes, we do not need temporary predictions for opfl modes to skip masked compound types as they are not used for opfl modes.
> 
> Further, by doing this, we are also reducing repeated opfl predictions. We have tested this runtime savings several times.

Yes, we do not need temporary predictions for opfl modes to skip masked compound types as they are not used for opfl modes.

Further, by doing this, we are also reducing repeated opfl predictions. We have tested this runtime savings several times.

However, since we have some early terminations based on cost, when we try to skip this process for all modes that do not support masked compound types, run time savings is not much.

yeqing-wu (approved):
LGTM

PR #5129, opened 2026-07-20
 Add 6 param warp search in winner mode by tid 

jadhuran:
Enable winner motion mode.

Based on tid threshold, choose to perform six parameter warp searches in either winner mode or rough mode.

Tests - cpu-used =1
Results (RA):

A1 - 17 frames
A2 - 33 frames

```
| A1      | -0.01% | 0.02%  | 0.08%  | -0.01% | 98.2%    | 100%     |
| A2      | 0.01%  | -0.07% | -0.01% | 0.01%  | 98.4%    | 101%     |
```

### DIFF
--- a/av2/encoder/rdopt.c
+++ b/av2/encoder/rdopt.c
@@ -1651,7 +1651,7 @@ static INLINE int select_modes_to_search(const AV2_COMP *const cpi,
   // the motion_mode_for_winner_cand speed feature, avoid searching it again.
   if (cpi->sf.winner_mode_sf.motion_mode_for_winner_cand) {
     if (!eval_motion_mode) {
-      modes_to_search = (1 << SIMPLE_TRANSLATION);
+      modes_to_search |= (1 << SIMPLE_TRANSLATION);
     } else {
       // Skip translation, as will have already been evaluated
       modes_to_search &= ~(1 << SIMPLE_TRANSLATION);
@@ -2533,7 +2533,8 @@ static AVM_INLINE int handle_warp_delta_mode(
     MB_MODE_INFO_EXT *mbmi_ext, HandleInterModeArgs *args, int64_t ref_best_rd,
     const BUFFER_SET *orig_dst, int_mv *previous_mvs,
     warp_mode_info_array *prev_best_models, int org_warp_inter_intra,
-    int rate2_nocoeff, int rate_mv0, int *tmp_rate_mv, int *tmp_rate2) {
+    int rate2_nocoeff, int rate_mv0, int *tmp_rate_mv, int *tmp_rate2,
+    int eval_motion_mode) {
   const AV2_COMMON *cm = &cpi->common;
   MACROBLOCKD *xd = &x->e_mbd;
   const int mi_row = xd->mi_row;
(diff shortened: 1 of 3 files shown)


av2/encoder/rdopt.c
```diff
@@ -8327,8 +8343,16 @@ static void tx_search_best_inter_candidates(
 }
 
 // Indicates number of winner simple translation modes to be used
-static const unsigned int num_winner_motion_modes[3] = { 0, 10, 3 };
-
+static const unsigned int num_winner_motion_modes[3] = { 0, 10, 6 };
+
+// Returns true if a mode is added to the winner mode check.
+static AVM_INLINE int is_check_in_winner(AV2_COMP *cpi,
+                                         MB_MODE_INFO *const mbmi) {
```
yeqing-wu: is_check_in_winner params should be const — reads only. Thus it is better to add const for both of parameters

av2/encoder/speed_features.c
```diff
@@ -359,6 +359,12 @@ static void set_good_speed_features_framesize_independent(
     // Cap the DRL depth for a fresh single-ref NEWMV search; reuse the
     // nearest searched result beyond the cap.
     sf->mv_sf.newmv_drl_search_limit = 2;
+    sf->part_sf.enable_six_param_warp_in_winner_mode = 1;
+    if (sf->part_sf.enable_six_param_warp_in_winner_mode) {
```
yeqing-wu: Is this condition truly necessary?

av2/encoder/speed_features.h
```diff
@@ -422,6 +422,10 @@ typedef struct PARTITION_SPEED_FEATURES {
   // the current best partition's boundary after searching NONE, HORZ, VERT, and
   // H-parts.
   int prune_part_4_with_partition_boundary;
+  // Enable six parameter warp in winner mode by tid. If set to 1, enable six
+  // parameter warp in winner mode by tid threshold.
+  int enable_six_param_warp_in_winner_mode;
+  int enable_six_param_warp_in_winner_mode_by_tid;
```
yeqing-wu: Since these are not partition speed features, it’s better to move them out of PARTITION_SPEED_FEATURES and into INTRA_MODE_SPEED_FEATURES or WINNER_MODE_SPEED_FEATURES.

av2/encoder/speed_features.h
```diff
@@ -422,6 +422,10 @@ typedef struct PARTITION_SPEED_FEATURES {
   // the current best partition's boundary after searching NONE, HORZ, VERT, and
   // H-parts.
   int prune_part_4_with_partition_boundary;
+  // Enable six parameter warp in winner mode by tid. If set to 1, enable six
+  // parameter warp in winner mode by tid threshold.
```
yeqing-wu: The comment is a bit misleading. I believe this comment is only relevant to enable_six_param_warp_in_winner_mode_by_tid, so we miss the comment for enable_six_param_warp_in_winner_mode, right?

av2/encoder/rdopt.c
```diff
@@ -8327,8 +8343,16 @@ static void tx_search_best_inter_candidates(
...
+  if (!cpi->sf.part_sf.enable_six_param_warp_in_winner_mode) return 0;
+  PREDICTION_MODE this_mode = mbmi->mode;
+  if (this_mode == WARP_NEWMV) return 1;
+  return 0;
+}
```
yeqing-wu, line 8383: Add a blank line?

av2/encoder/rdopt.c
```diff
@@ -2579,6 +2580,8 @@ static AVM_INLINE int handle_warp_delta_mode(
   if (!allow_warp_parameter_signaling(cm, mbmi)) {
     if (mbmi->warp_precision_idx) return -1;
 
+    // Default parameters are not searched if the delta is not
+    // signalled
```
yeqing-wu: Is this comment misplaced?

av2/encoder/speed_features.c
jadhuran: Yes, since we define tid based feature associated to this.

av2/encoder/rdopt.c
jadhuran: I will address

av2/encoder/speed_features.h
jadhuran: I will address

av2/encoder/speed_features.h
jadhuran: ok, I will have separate comments for each feature

av2/encoder/rdopt.c
jadhuran: I will address

av2/encoder/rdopt.c
jadhuran: No, it has been incorrectly added. Now removed

av2/encoder/rdopt.c
jadhuran: Addressed

av2/encoder/speed_features.c
jadhuran: Addressed

av2/encoder/speed_features.h
jadhuran: Addressed

av2/encoder/speed_features.h
jadhuran: Addressed

av2/encoder/rdopt.c
jadhuran: Addressed

av2/encoder/speed_features.c
yeqing-wu: Sorry, my question is: we assign 1 to sf->part_sf.enable_six_param_warp_in_winner_mode, then we immediately check `if (sf->part_sf.enable_six_param_warp_in_winner_mode)`, thus it is always true, this condition is not necessary.

jadhuran: Since, this is a speed feature, when we turn it off, everything related to it will be turned off. For example, the speed feature will never be turned on if this is turned off. It is placed, so that we control overall changes with one feature. Second feature sets tid level. We also want to control outside major functions.

yeqing-wu: To achieve that goal, I think you should move the following code outside the speed >= 1 block. That way, motion_mode_for_winner_cand and enable_six_param_warp_in_winner_mode_by_tid aren't determined by the speed preset — they're determined by enable_six_param_warp_in_winner_mode, and only enable_six_param_warp_in_winner_mode is determined by the speed preset. As the code stands now, the check is redundant.

```
    if (sf->inter_sf.enable_six_param_warp_in_winner_mode) {
      sf->winner_mode_sf.motion_mode_for_winner_cand = 2;
      sf->inter_sf.enable_six_param_warp_in_winner_mode_by_tid =
          cm->current_frame.pyramid_level >= 3 ? 1 : 0;
    }
```

jadhuran: Ok, I understand your point now. I have moved that code block outside presets as you suggested.

yeqing-wu (approved):

leolzhao (approved):

PR #5130, opened 2026-07-20
Add a speed level for intra region in inter frames

leolzhao:
Squash merge branch 'leo/inter-sdp-mr' into 'avm-encoder-development' Add a speed level for inter sdp fast method:

1) Early terminate inter-sdp when there is no intra coded blocks in the this region after searching mixed intra and inter region;
2) Early terminate inter-sdp when more than half of the blocks are inter -coded;
3) Early terminate inter-sdp when the best block partitioning is partition-none

STATS_CHANGED

cpu-use-1 is at commit 6e71b0c859 is used as anchor.

A1 - 17 frames (0.02% loss with 95.3% run-time)
A2 - 33 frames (0.09% loss with 94.6% run-time)

```
| A1      | 0.01% | 0.17%  | 0.11% | 0.02% | 95.3%    | 100%   |
| A2      | 0.07% | 0.14%  | 0.42% | 0.09% | 94.6%    | 100%   |
```

### DIFF
--- a/av2/encoder/partition_search.c
+++ b/av2/encoder/partition_search.c
@@ -4496,8 +4496,8 @@ static AVM_INLINE void prune_ext_partitions_3way(
 
 // Early termination of SDP for intra blocks in inter frames
 static INLINE void early_termination_inter_sdp(PC_TREE *pc_tree,
-                                               float *total_count,
-                                               float *inter_mode_count) {
+                                               unsigned int *total_count,
+                                               unsigned int *inter_mode_count) {
   REGION_TYPE cur_region_type = pc_tree->region_type;
   if (pc_tree->partitioning == PARTITION_NONE) {
     if (pc_tree->none[cur_region_type] == NULL) return;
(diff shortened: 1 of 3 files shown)


av2/encoder/partition_search.c
```diff
@@ -4599,17 +4599,29 @@ static INLINE void search_intra_region_partitioning(
...
   float total_count = 0;
   float inter_mode_count = 0;
   early_termination_inter_sdp(pc_tree, &total_count, &inter_mode_count);
-  // if over 60% of the coded blocks under this region are inter coded blocks,
-  // skip the rdo for inter-sdp
-  if (total_count * 0.7 < inter_mode_count) return;
+  if (cpi->sf.part_sf.inter_sdp_fast_method_level >= 1) {
+    // Optimized fast method (enabled at cpu-used >= 1):
+    // Check whether there are at least 1 intra coded blocks under this region
+    if (total_count - inter_mode_count < 1) return;
```
yunqingwang1, line 4612: Can both total_count and inter_mode_count be 0? Maybe add an extra check for that?

yunqingwang1:
Nice speedup and tradeoff! Looks good overall.

av2/encoder/partition_search.c
```diff
@@ -4599,17 +4599,29 @@ static INLINE void search_intra_region_partitioning(
   MACROBLOCKD *const xd = &x->e_mbd;
   assert(pc_tree != NULL);
 
+  // store the current best partitioning
+  PARTITION_TYPE cur_best_partitioning = pc_tree->partitioning;
+
   // Add one encoder fast method for early terminating inter-sdp
   float total_count = 0;
   float inter_mode_count = 0;
   early_termination_inter_sdp(pc_tree, &total_count, &inter_mode_count);
```
yeqing-wu, line 4608: Just curious (not because of this PR) about the reason behind using float type for total_count and inter_mode_count. Shouldn’t it be better to use int, as it only increases the integer value for both of them inside early_termination_inter_sdp?

yeqing-wu (approved):
LGTM

yeqing-wu:

av2/encoder/partition_search.c
leolzhao: Good catch! Just changed the variable to unsigned int to address this issue.

av2/encoder/partition_search.c
leolzhao: Just changed the variable from float to unsigned int to avoid confusion.

PR #5132, opened 2026-07-20
Increase TCQ threshold based on base QP

motongxu:
Increase TCQ zero threshold to 9 * abs_coef >= 10 * dequant when base_qindex < 185.

Tests: cpu-used =1, RA
A1 - 17 frames
A2 - 33 frames

```
| A1      | -0.01% | -0.01% |  0.00% | -0.01% | 98.6%    | 98.6%    |
| A2      |  0.00% | -0.06% | -0.05% |  0.00% | 98.9%    | 100.1%   |
```


yeqing-wu:

av2/common/enums.h
```diff
@@ -30,6 +30,8 @@ extern "C" {
 
 #undef MAX_SB_SIZE
 
+#define MAX_TCQ_THRES_QIDX 185
```
yeqing-wu: MAX_TCQ_THRES_QIDX is placed in this header file but is only used by the encoder. The decoder doesn’t rely on it, so placing an encoder tuning constant in a common header file can pollute decoder builds. It’s better to move it to av2/encoder/speed_features.h or av2_quantize.h.

av2/encoder/av2_quantize.c
```diff
@@ -50,7 +50,12 @@ static void highbd_quantize_fp_helper_c(
   (void)zbin_ptr;
   (void)quant_shift_ptr;
   (void)iscan;
-
+  int abs_coeff_scale = 1;
+  int dequant_ptr_scale = 1;
+  if (deadzone_thres == 1) {  // 5/9
```
yeqing-wu: Is this comment incorrect? Should it be 10/9 instead?

av2/encoder/x86/av2_highbd_quantize_avx2.c
```diff
@@ -65,12 +65,20 @@ static AVM_FORCE_INLINE __m256i get_max_lane_eob(const int16_t *iscan_ptr,
 static INLINE void quantize(const __m256i *qp, const tran_low_t *coeff_ptr,
                             const int16_t *iscan_ptr, int log_scale,
                             tran_low_t *qcoeff, tran_low_t *dqcoeff,
-                            __m256i *eob) {
+                            __m256i *eob, int deadzone_thres) {
```
yeqing-wu: Is deadzone_thres a threshold or just a boolean indicating whether to use the tcq deadzone or not? If it’s later, should we rename it to use_tcq_deadzone_boost to avoid confusion? The same applies to all places that use deadzone_thres.

av2/encoder/encodemb.c
```diff
@@ -949,6 +941,19 @@ static void encode_block(int plane, int block, int blk_row, int blk_col,
                             INTER_BLOCK_OPT_TYPE == TRELLIS_DROPOUT_OPT;
     const bool do_dropout = INTER_BLOCK_OPT_TYPE == DROPOUT_OPT ||
                             INTER_BLOCK_OPT_TYPE == TRELLIS_DROPOUT_OPT;
+    const int deadzone_thres =
+        use_tcq && quant_param.use_optimize_b && do_trellis && !fsc_mode &&
+        cpi->sf.tx_sf.enable_adaptive_tcq_threshold &&
+        cm->quant_params.base_qindex < MAX_TCQ_THRES_QIDX;
```
yeqing-wu: tx_search.c:2477-2480 uses deadzone_thres without requiring do_trellis. However, 944-946 and 1455-1457 require it in this file. Is this expected? Should we align the predicate at both sites?

```
      const int deadzone_thres =
          use_tcq && quant_param_intra.use_optimize_b && !fsc_mode &&
          cpi->sf.tx_sf.enable_adaptive_tcq_threshold &&
          cm->quant_params.base_qindex < MAX_TCQ_THRES_QIDX;
```

avm_dsp/quantize.h
```diff
@@ -13,6 +13,8 @@
 #ifndef AVM_AVM_DSP_QUANTIZE_H_
 #define AVM_AVM_DSP_QUANTIZE_H_
 
+#include <stdbool.h>
```
yeqing-wu: Do we really need this header file? If not, please remove it.

avm_dsp/x86/highbd_quantize_intrin_sse2.c
```diff
@@ -11,6 +11,7 @@
  */
 
 #include <emmintrin.h>
+#include <stdbool.h>
```
yeqing-wu, line 14: Same comment, do we really need this header file? If not, please remove it.

avm_dsp/quantize.c
```diff
@@ -201,14 +201,15 @@ void avm_highbd_quantize_b_64x64_adaptive_c(
...
-                             const int32_t *quant_ptr,
+void avm_highbd_quantize_b_c(int deadzone_thres, const tran_low_t *coeff_ptr,
+                             intptr_t n_coeffs, const int32_t *zbin_ptr,
+                             const int32_t *round_ptr, const int32_t *quant_ptr,
                              const int32_t *quant_shift_ptr,
                              tran_low_t *qcoeff_ptr, tran_low_t *dqcoeff_ptr,
                              const int32_t *dequant_ptr, uint16_t *eob_ptr,
                              const int16_t *scan, const int16_t *iscan,
                              const int log_scale) {
+  (void)deadzone_thres;
```
yeqing-wu: deadzone_thres is unused in the DSP layer and both DSP SIMD files — (void)deadzone_thres).
Thus, it is better to revert the DSP RTCD signature change

test/av2_quantize_test.cc
```diff
@@ -175,12 +175,12 @@ class AV2QuantizeTest : public ::testing::TestWithParam<QuantizeFuncParams> {
...
       }
 
-      quanFuncRef(coeff_ptr, count, zbin_ptr, round_ptr, quant_ptr,
+      quanFuncRef(0, coeff_ptr, count, zbin_ptr, round_ptr, quant_ptr,
                   quant_shift_ptr, ref_qcoeff_ptr, ref_dqcoeff_ptr, dequant_ptr,
                   &ref_eob, scanOrder.scan, scanOrder.iscan, log_scale);
 
       ASM_REGISTER_STATE_CHECK(
-          quanFunc(coeff_ptr, count, zbin_ptr, round_ptr, quant_ptr,
+          quanFunc(0, coeff_ptr, count, zbin_ptr, round_ptr, quant_ptr,
```
yeqing-wu: In the unit tests, the code forces the use of a deadzone_thres of 0. It would be better to have unit tests that cover the scenario where deadzone_thres is 1. Otherwise, we won’t know if the code is correct or not.

av2/common/enums.h
motongxu: Moved this macro to speed feature instead. Thanks for suggestion.

av2/encoder/av2_quantize.c
motongxu: The original threshold was 1/2 according to the following if conditions (<<-1 and <<1):
```
if (abs_coeff_scale *
              (((tran_high_t)abs_coeff * wt) << QUANT_TABLE_BITS) >=
          dequant_ptr_scale * ((tran_high_t)dequant_ptr[rc != 0]
                               << (AVM_QM_BITS - (1 + log_scale))))
```
and
```
if (abs_coeff_scale *
              ((tran_high_t)abs_coeff << (1 + log_scale + QUANT_TABLE_BITS)) >=
          dequant_ptr_scale * (tran_high_t)dequant_ptr[rc01])
```
So, the actual threshold here is 10/9.

av2/encoder/x86/av2_highbd_quantize_avx2.c
motongxu: The names are all updated to "use_tcq_deadzone_boost".

av2/encoder/encodemb.c
motongxu: Yes, this is expected.
The predicate for use_tcq_deadzone_boost should match the condition under which av2_optimize_b() is called. In encodemb.c, the av2_optimize_b() call is guarded by do_trellis, so use_tcq_deadzone_boost also includes that condition. In tx_search.c, however, the corresponding av2_optimize_b() call is not guarded by do_trellis.

avm_dsp/quantize.h
motongxu: This has been removed. Thanks for pointing out.

avm_dsp/x86/highbd_quantize_intrin_sse2.c
motongxu: This has also been removed.

test/av2_quantize_test.cc
motongxu: Test cases are added with use_tcq_deadzone_boost=1.

avm_dsp/quantize.c
motongxu: av2_highbd_quantize_* and avm_highbd_quantize_* share the same RTCD dispatch table and the same “typedef (QuantizeFpFunc)”, so they must have identical signatures. So what I did here is to keep function signature the same and use (void) to ignore it for avm_ functions.

yeqing-wu (approved):
LGTM

PR #5133, opened 2026-07-20
Enhanced inter mode cache reuse

leolzhao:
Reuse model RD when one of the following conditions are met.

- Current mode is the same as the cached mode and at least one of the reference frame is the same as the cached one

- All the reference frames is the same as the reference frame of the cached mode

Otherwise, skip processing current mode or reference frame

STATS_CHANGED

Results are tested on top of 6e71b0c85924e61d9 with cpu-used = 1.
```
| A1      | 0.03% | -0.05%  | 0.00% | 0.03% | 97.8%    | 100%   |
| A2      | 0.03% | 0.41%   | 0.13% | 0.05% | 96.6%    | 100%   |
```

### DIFF
--- a/av2/encoder/rdopt.c
+++ b/av2/encoder/rdopt.c
@@ -7491,12 +7491,30 @@ static INLINE int is_mode_intra(PREDICTION_MODE mode) {
   return mode < INTRA_MODE_END;
 }
 
+// Check whether one of the reference frames are the same as that of the cached
+// mode
+static INLINE int is_ref_frame_same_as_cache(
+    const MB_MODE_INFO *cached_mi2, const MV_REFERENCE_FRAME *ref_frame,
+    int this_mode_is_single) {
+  const int cached_mode_is_single = is_inter_singleref_mode(cached_mi2->mode);
+  if (ref_frame[0] == cached_mi2->ref_frame[0]) return 1;
+  if (!cached_mode_is_single && (ref_frame[0] == cached_mi2->ref_frame[1]))
+    return 1;
+  if (!this_mode_is_single) {
+    if (ref_frame[1] == cached_mi2->ref_frame[0]) return 1;
+    if (!cached_mode_is_single && (ref_frame[1] == cached_mi2->ref_frame[1]))
+      return 1;
+  }
+  return 0;
+}
+
 // Reuse the prediction mode in cache.
 // Returns 0 if no pruning is done, 1 if we are skipping the current mod
 // completely, 2 if we skip compound only, but still try single motion modes
 static INLINE int skip_inter_mode_with_cached_mode(
-    const AV2_COMMON *cm, const MACROBLOCK *x, PREDICTION_MODE mode,
+    const AV2_COMP *cpi, const MACROBLOCK *x, PREDICTION_MODE mode,
     const MV_REFERENCE_FRAME *ref_frame) {
+  const AV2_COMMON *const cm = &cpi->common;
   const MB_MODE_INFO *cached_mi = x->inter_mode_cache[0];
 
   // If there is no cache, then no pruning is possible.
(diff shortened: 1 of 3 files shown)


av2/encoder/rdopt.c
```diff
@@ -7466,19 +7467,45 @@ static INLINE int skip_inter_mode_with_cached_mode(
...
-            (ref_frame[0] == cached_mi2->ref_frame[1]))
-          return 0;
-        if (!this_mode_is_single) {
-          if (ref_frame[1] == cached_mi2->ref_frame[0]) return 0;
+        if (cpi->sf.inter_sf.enable_enhanced_inter_mode_cache_reuse) {
+          // Enhanced cache reuse:
+          // For single prediction mode, when reference frame is the same,
+          // return 0 For compound prediction mode, when both reference frames
+          // are the same, return 0 When prediction mode is the same, when one
+          // of the refernce frames are the same, return 0
```
yeqing-wu: This comment appears to be malformed, possibly due to a missing comma or period. Could you please verify its correctness?

av2/encoder/rdopt.c
```diff
@@ -7466,19 +7467,45 @@ static INLINE int skip_inter_mode_with_cached_mode(
...
+          if (cached_mode_is_single && ref_frame[0] == cached_mi2->ref_frame[0])
+            return 0;
+          else {
+            if (ref_frame[0] == cached_mi2->ref_frame[0] &&
+                ref_frame[1] == cached_mi2->ref_frame[1])
+              return 0;
+          }
```
yeqing-wu, line 7545: Should these codes actually be presented as shown in the following? With the current logic, if it’s a single frame, and cached_mode_is_single is true but ref_frame[0] != cached_mi2->ref_frame[0], the code would go to the else branch and have to perform another check. Is that unnecessary?

```
  if (cached_mode_is_single) {
    if (ref_frame[0] == cached_mi2->ref_frame[0]) return 0;
  } else {
    if (ref_frame[0] == cached_mi2->ref_frame[0] &&
        ref_frame[1] == cached_mi2->ref_frame[1]) return 0;
  }
```

av2/encoder/rdopt.c
yeqing-wu: and typo 'refernce'

av2/encoder/rdopt.c
```diff
@@ -7466,19 +7467,45 @@ static INLINE int skip_inter_mode_with_cached_mode(
...
+            if (ref_frame[0] == cached_mi2->ref_frame[0]) return 0;
+            if (!cached_mode_is_single &&
+                (ref_frame[0] == cached_mi2->ref_frame[1]))
+              return 0;
+            if (!this_mode_is_single) {
+              if (ref_frame[1] == cached_mi2->ref_frame[0]) return 0;
+              if (!cached_mode_is_single &&
+                  (ref_frame[1] == cached_mi2->ref_frame[1]))
+                return 0;
+            }
```
yeqing-wu: All the code inside the `if (mode == cached_mi2->mode) { }` block is identical to the branch of the original cache reuse logic. Can we eliminate this?

By the way, do we still need the older logic? Have you tested whether your new logic can completely replace the old one? What would happen in such a case?

yeqing-wu:

av2/encoder/rdopt.c
leolzhao: Yes, your suggested changes are more reasonable.

av2/encoder/rdopt.c
leolzhao: Done.

av2/encoder/rdopt.c
leolzhao: Modified the code as you suggested.

av2/encoder/rdopt.c
leolzhao: Added one function to move the common part into that function.

The new logic will cause around 0.06% loss for cpu-used = 0, which is not small. So, we still need the old logic for cpu-used = 0.

leolzhao:
Please help review/approve this MR again, thanks! @yeqing-wu @yunqingwang1

yeqing-wu (approved):
LGTM

PR #5144, opened 2026-07-22
Enable skip tx search when max_eob<512

motongxu:
Skip remaining tx type search for small blocks when best eob is 0.

Tests: cpu-used =1, RA
A1 - 17 frames
A2 - 33 frames

```
| A1      |  0.09% |  0.20% |  0.04% |  0.09% | 96.9%    | 99.0%    |
| A2      | -0.02% |  0.04% | -0.49% | -0.04% | 97.0%    | 100.0%   |
```


av2/encoder/speed_features.c
```diff
@@ -808,6 +813,8 @@ static AVM_INLINE void init_tx_sf(TX_SPEED_FEATURES *tx_sf) {
   tx_sf->tx_type_search.use_reduced_intra_txset = 0;
   tx_sf->tx_type_search.fast_inter_tx_type_search = 0;
   tx_sf->tx_type_search.skip_tx_search = 0;
+  tx_sf->tx_type_search.blk_size_adapt_skip_tx_search = false;
+  tx_sf->tx_type_search.skip_tx_search_max_blk_size = 1024;
```
yeqing-wu: I think the naming is misleading — we don't have block sizes > 256. Based on the code, it looks like the logic is: if the eob is larger than a certain value, we skip the tx search, right? If so, should we rename it to skip_tx_search_max_eob? 
Same question for blk_size_adapt_skip_tx_search — should it be eob_adapt_skip_tx_search?

yeqing-wu: BTW, this tradeoff for class A2 is very nice.

motongxu: Yest, you are right. I have updated the naming accordingly. Thank you.

yeqing-wu (approved):
LGTM, thanks

PR #5149, opened 2026-07-24
Speed up on uneven 4way partition and MV precision.

jadhuran:
This has two contributions

Contribution 1

Block size based partition depth reduction for uneven 4 way

Q index threshold is applied only to key frames.

Apply separate partition depth for uneven 4 way partition from H   - partition

Contribution 2

Prune mv using best precision

Do not prune one pel when maximum precision is quarter pel.

Results:
Anchor: 1d2febc43d7dd206e68ac178cf7e69e17652d0dd

Test: Contribution 1 + Contribution 2

```
| Summary |   Y    |   U   |   V   |  YUV  | Enc-time | Dec-time |
| A1      | -0.02% | 0.24% | 0.21% | 0.00% | 97.8%    | 100.5%   |
| A2      | 0.10%  | 0.15% | 0.03% | 0.09% | 95.8%    | 99.8%    |
```
Test: Contribution 1

```
+---------+--------+-------+--------+--------+----------+----------+
| Summary |   Y    |   U   |   V    |  YUV   | Enc-time | Dec-time |
+---------+--------+-------+--------+--------+----------+----------+
| A1      | -0.02% | 0.05% | -0.14% | -0.02% | 99.0%    | 100.5%   |
| A2      | 0.06%  | 0.09% | 0.19%  | 0.07%  | 97.6%    | 99.3%    |
+---------+--------+-------+--------+--------+----------+----------+

```


av2/encoder/rdopt.h
```diff
@@ -374,13 +374,24 @@ static INLINE void update_mv_precision(const MV ref_mv,
...
+
   if (!prune_mv_prec_using_best_mv_prec_so_far) return false;
   if (prune_mv_prec_using_best_mv_prec_so_far >= 1) {
     // If the current MV precision index is farther from the best MV precision,
     // prune the evaluation of current MV precision.
+
+    // Avoid pruning one pel precision for boosted frames based
+    // on qindex when mbmi->max_mv_precision is QTR_PEL.
+    if (cpi->sf.flexmv_sf.prune_non_one_pel_mv_using_best_mv_prec &&
+        mbmi->max_mv_precision == MV_PRECISION_QTR_PEL && precision_dx == 2)
```
yeqing-wu: The flag name `prune_non_one_pel_mv_using_best_mv_prec` promises "restrict pruning one pel" universally, but ONE_PEL is only preserved when max_mv_precision == QTR_PEL. Other max precisions where ONE_PEL is in the set (ONE_EIGHTH_PEL: set 0, dx=1; HALF_PEL: set 2, dx=2) get no exemption, and future precision-set edits silently break the check. should we use precision_def->precision[precision_dx] == MV_PRECISION_ONE_PEL to replace precision_dx == 2?

av2/encoder/rdopt.h
```diff
@@ -374,13 +374,24 @@ static INLINE void update_mv_precision(const MV ref_mv,
 }
 // Prune the evaluation of current MV precision based on best MV precision
 // chosen so far.
-static INLINE bool prune_curr_mv_precision_eval(
-    int prune_mv_prec_using_best_mv_prec_so_far, int precision_dx,
-    int best_precision_dx) {
+static INLINE bool prune_curr_mv_precision_eval(AV2_COMP *const cpi,
+                                                const MB_MODE_INFO *mbmi,
```
yeqing-wu: Since we don’t change the values of the items, we should add constants for both the value and pointer for cpi and mbmi.

av2/encoder/speed_features.h
```diff
@@ -981,6 +983,9 @@ typedef struct FLEXMV_PRECISION_SPEED_FEATURES {
   // Prune the evaluation of current MV precision based on best MV precision
   // chosen so far.
   int prune_mv_prec_using_best_mv_prec_so_far;
+  // Enable prune mv precision using best mv precision but restrict pruning one
+  // pel
+  int prune_non_one_pel_mv_using_best_mv_prec;
```
yeqing-wu, line 990: Could you please clarify whether we should enable or disable the screen content tool, such as Intrabc, for a speed of 1 or higher?

av2/encoder/rdopt.h
```diff
@@ -374,13 +374,24 @@ static INLINE void update_mv_precision(const MV ref_mv,
...
+                                                int precision_dx,
+                                                int best_precision_dx) {
+  const int prune_mv_prec_using_best_mv_prec_so_far =
+      cpi->sf.flexmv_sf.prune_non_one_pel_mv_using_best_mv_prec ||
+      cpi->sf.flexmv_sf.prune_mv_prec_using_best_mv_prec_so_far;
+
   if (!prune_mv_prec_using_best_mv_prec_so_far) return false;
   if (prune_mv_prec_using_best_mv_prec_so_far >= 1) {
```
yeqing-wu: The code after the change is dead code because prune_mv_prec_using_best_mv_prec_so_far can only be 0 or 1. If it is 0, the code above it already returns. Therefore, the condition check is dead code.

`if (prune_mv_prec_using_best_mv_prec_so_far >= 1) {`

av2/encoder/rdopt.h
```diff
@@ -374,13 +374,24 @@ static INLINE void update_mv_precision(const MV ref_mv,
...
 // Prune the evaluation of current MV precision based on best MV precision
 // chosen so far.
-static INLINE bool prune_curr_mv_precision_eval(
-    int prune_mv_prec_using_best_mv_prec_so_far, int precision_dx,
-    int best_precision_dx) {
+static INLINE bool prune_curr_mv_precision_eval(AV2_COMP *const cpi,
+                                                const MB_MODE_INFO *mbmi,
+                                                int precision_dx,
+                                                int best_precision_dx) {
+  const int prune_mv_prec_using_best_mv_prec_so_far =
```
yeqing-wu: The name “prune_mv_prec_using_best_mv_prec_so_far” is misleading because its value depends on both “sf.flexmv_sf.prune_non_one_pel_mv_using_best_mv_prec” and “sf.flexmv_sf.prune_mv_prec_using_best_mv_prec_so_far.” However, the code name for this local variable originates from one of the speed features.

av2/encoder/speed_features.h
```diff
@@ -410,6 +410,8 @@ typedef struct PARTITION_SPEED_FEATURES {
   // If set to 1, recursion depth is set 2 if block area > 1024 else it is set
   // to INT_MAX. If set to 2, recursion depth is set to 1.
   int ext_recur_depth_level;
+  // reduce the recursion depth for uneven 4-way partitions based on block size
```
yeqing-wu: Comment style break - lowercase, no period.

av2/encoder/speed_features.h
jadhuran: In comment? Why disabling intrabc matters here. This is applied for non-intrabc inter modes.

av2/encoder/rdopt.h
jadhuran: No, QTR_PEL being max precision is important, because we haven't tested other sets without this condition. If set changes, this will impact.  prune_mv_using_best_mv_prec, does this sound better?

av2/encoder/rdopt.h
jadhuran: This is an existing code, but I can remove it.

av2/encoder/rdopt.h
jadhuran: ok

av2/encoder/rdopt.h
jadhuran: prune_mv_prec_using_best_mv_prec_so_far this is an existing feature. Shall I name my new feature to be prune_mv_using_best_mv_prec?

av2/encoder/speed_features.h
jadhuran: Addressed

av2/encoder/rdopt.h
jadhuran: Addressed

av2/encoder/rdopt.h
jadhuran: Addressed

av2/encoder/rdopt.h
yeqing-wu: You still keep the max precision mbmi->max_mv_precision == MV_PRECISION_QTR_PEL. What I mean is to replace precision_dx == 2 with precision_def->precision[precision_dx] == MV_PRECISION_ONE_PEL, so even you change the max precision, you still have the correct logic and do not need to change '2'. Meanwhile it is also easy to understand than magic number '2'.

av2/encoder/speed_features.h
yeqing-wu: I just want to check if this would impact intrabc or not since in AV2, I think it also supports sub-pel. Since you said it is for non-intrabc inter modes, then you answered my question.

av2/encoder/rdopt.h
yeqing-wu: I know prune_mv_prec_using_best_mv_prec_so_far is an existing feature, but here you're repurposing it for two features while still using prune_mv_prec_using_best_mv_prec_so_far as the local variable name — and there's also a prune_mv_prec_using_best_mv_prec_so_far coming from flexmv_sf. That's misleading.
prune_mv_using_best_mv_prec looks good to me.

yeqing-wu (approved):

av2/encoder/rdopt.h
jadhuran: I understand your point. Addressed.

av2/encoder/rdopt.h
jadhuran: Addressed

PR #5150, opened 2026-07-24
Remove encoder-side limitation for restricted switch frames

### LINKED ISSUE #5152: Encoder limits prediction modes for restricted switch frames
### What version / commit were you testing with?

Current main branch commit 636fcda84b06650b6b61493b4bf7337a9b102235

### What steps will reproduce the problem?

Encode the same bitstream with restricted and again with non-restricted switch frames in random access configuration. The rate of the switch frames is expect to be close to each other in both cases. The rate of the non-restricted switch frame is however much lower than the rate of the restricted switch frame. 

### Additional information

The function `prune_ref_by_selective_ref_frame()` returns 1 for all blocks that reference restricted references:
```
static INLINE int prune_ref_by_selective_ref_frame(
    const AV2_COMP *const cpi, const MACROBLOCK *const x,
    const MV_REFERENCE_FRAME *const ref_frame) {
  (void)x;
  const AV2_COMMON *const cm = &cpi->common;
  const SPEED_FEATURES *const sf = &cpi->sf;

  if (!sf->inter_sf.selective_ref_frame) return 0;
  assert(ref_frame[0] != NONE_FRAME);
  if (ref_frame[0] == INTRA_FRAME) return 0;

  RefCntBuffer *ref0_buf = get_ref_frame_buf(cm, ref_frame[0]);
  if (ref0_buf)
    if (ref0_buf->is_restricted) return 1;
  RefCntBuffer *ref1_buf = get_ref_frame_buf(cm, ref_frame[1]);
  if (ref1_buf)
    if (ref1_buf->is_restricted) return 1;
```
This causes `WARPMV` to be the only prediction mode to be considered for restricted references by the encoder, severely hindering the compression performance of restricted switch frames and post-switch frames referencing restricted references.

mabenjak:
The function prune_ref_by_selective_ref_frame() returned 1 for all blocks that reference restricted references. This caused WARPMV to be the only prediction mode to be considered for restricted references by the encoder, severely hindering the compression performance of restricted switch frames. Removed this limitation and set display_order_hint to REF_RESTRICTED_DOH to ensure that all now accessible prediction modes are correctly encoded. Added is_restricted checks to av2_get_refresh_frame_flags() because the display_order_hint of restricted references is always >= the current display_order_hint.

Fixes #5152


urvangjoshi:
@mabenjak : is this related to ongoing VCWG discussion?

PR #5155, opened 2026-07-24
Add default speed features for RTC

jianj-g:


yunqingwang1:
@leolzhao @yeqing-wu  Jerome adds some speed features for real-time encoding. We didn't define FG16 CTC for RTC case. Would LD CTC be enough? Any thoughts?

av2/encoder/speed_features.h
```diff
@@ -957,6 +960,23 @@ typedef struct LOOP_FILTER_SPEED_FEATURES {
...
+  bool use_first_reference_only;
+
+  // Flag to disable all but DC intra mode for inter frame prediction.
+  bool use_only_dc_intra_interframe;
+
+  // Disable GDF.
+  int disable_gdf;
+
+  // Disable CCSO.
+  int disable_ccso;
```
yeqing-wu: We already have the command line options (which are available in the encoder section) to enable and disable:
```
use_only_dc_intra_interframe-->use_inter_dct_only
disable_ccso-->enable_ccso
disable_gdf-->enable_cdef
```

Can we achieve the goal of using only the first reference by setting `max_reference_frames` to 1?

Since the existing flags in the code already accomplish this, why do we introduce these new features? Introducing more duplicated flags would make the code harder to understand and confusing for users regarding their configuration.

jianj-g: Removed `disable_csso`, `disable_gdf` and `use_first_reference`

`use_only_dc_intra_interframe` means using only DC for intra modes in inter frames, which is different from `use_inter_dct_only`.

yeqing-wu: Thanks for the changes. You’re right; it’s DC, not DCT. I misread it as DCT.

yeqing-wu (approved):

PR #5156, opened 2026-07-24
Add frame level on/off bawp

jianle973:
Enable frame level bawp based on histogram. Enabled for speed 2 and above.

33 frame RA results for speed 2 and 3 are as follows: 

                                                    | Summary | Y | U | V | YUV | Enc-time | Dec-time |
Speed2, Overall w/o B2 (RA)    +0.00%   +0.01%   +0.01%   +0.00%  99.51%  100.12%
Speed3, Overall w/o B2 (RA)    +0.00%   +0.01%   +0.03%   +0.00%  99.45%   99.53%

### DIFF
--- a/av2/encoder/encodeframe.c
+++ b/av2/encoder/encodeframe.c
@@ -1870,6 +1870,75 @@ static AVM_INLINE void av2_enc_setup_tip_frame(AV2_COMP *cpi) {
   }
 }
 
+static void get_histogram(int32_t *hist, const YV12_BUFFER_CONFIG *buf,
+                          int width, int height) {
+  uint16_t *y_buf = buf->y_buffer;
+  int stride = buf->y_stride;
+  const int step = 2;
+  for (int i = 0; i < height; i += step) {
+    for (int j = 0; j < width; j += step) {
+      hist[y_buf[j]]++;
+    }
+    y_buf += stride;
+  }
+}
+
+static int av2_set_on_bawp_picture_level(AV2_COMP *const cpi) {
+  int enable_curr_pic_bawp = 0;
+  assert(cpi->sf.inter_sf.enable_fast_bawp == 1);
+  AV2_COMMON *const cm = &cpi->common;
+  const YV12_BUFFER_CONFIG *source = cpi->source;
+
+  const int num_values = 1 << cm->seq_params.bit_depth;
+
+  int32_t *curr_hist;
+  int32_t *ref_hist;
+  CHECK_MEM_ERROR(cm, curr_hist, avm_calloc(num_values, sizeof(int32_t)));
+  CHECK_MEM_ERROR(cm, ref_hist, avm_calloc(num_values, sizeof(int32_t)));
+
+  const int width = source->y_width;
+  const int height = source->y_height;
+  const int step = 2;
+  const int num_samples = width * height / (step * step);
+
+  //----- get histogram diff threshold -----
+  const double sample_thres =
+      cm->features.allow_screen_content_tools ? 0.075 : 0.1;
+  const int32_t hist_diff_thres = (int32_t)(sample_thres * num_samples);
+
+  // get histogram of the current picture
+  get_histogram(curr_hist, source, width, height);
+
+  for (MV_REFERENCE_FRAME rf = 0; rf < cm->ref_frames_info.num_total_refs;
+       ++rf) {
+    const RefCntBuffer *ref_buf = get_ref_frame_buf(cm, rf);
+    if (ref_buf != NULL) {
+      int ref_width = ref_buf->mi_cols * MI_SIZE;
+      int ref_height = ref_buf->mi_rows * MI_SIZE;
+
+      if (width == ref_width && height == ref_height) {
+        memset(ref_hist, 0, sizeof(int32_t) * num_values);
+        // get histogram of the reference picture
+        get_histogram(ref_hist, &ref_buf->buf, width, height);
+
+        // get SAD of delta histogram
+        int32_t diff_hist = 0;
+        for (int i = 0; i < num_values; i++) {
+          diff_hist += abs(curr_hist[i] - ref_hist[i]);
+        }
+
+        if (diff_hist > hist_diff_thres) {
+          enable_curr_pic_bawp = 1;
+          break;
+        }
+      }
+    }
+  }
+  avm_free(curr_hist);
+  avm_free(ref_hist);
+  return enable_curr_pic_bawp;
+}
+
 /*!\brief Set the lossless flags for a frame before encoding it
  *
  * \ingroup high_level_algo
@@ -2147,6 +2216,10 @@ static AVM_INLINE void encode_frame_internal(AV2_COMP *cpi) {
   start_timing(cpi, av2_compute_global_motion_time);
 #endif
   av2_compute_global_motion_facade(cpi);
+
+  if (features->enable_bawp && cpi->sf.inter_sf.enable_fast_bawp)
+    features->enable_bawp = av2_set_on_bawp_picture_level(cpi);
+
 #if CONFIG_COLLECT_COMPONENT_TIMING
   end_timing(cpi, av2_compute_global_motion_time);
 #endif
--- a/av2/encoder/speed_features.c
+++ b/av2/encoder/speed_features.c
@@ -490,6 +490,7 @@ static void set_good_speed_features_framesize_independent(
       // motion vector predictor diversity and mitigating coding loss.
       sf->mv_sf.newmv_drl_search_limit = 2;
     }
+    sf->inter_sf.enable_fast_bawp = 1;
 
     sf->intra_sf.prune_palette_search_level = 1;
 
@@ -932,6 +933,7 @@ static AVM_INLINE void init_inter_sf(INTER_MODE_SPEED_FEATURES *inter_sf) {
   inter_sf->enable_enhanced_inter_mode_cache_reuse = 0;
   inter_sf->prune_comp_mode_eval_using_est_rd = false;
   inter_sf->prune_warp_newmv_ref_mv_idx = false;
+  inter_sf->enable_fast_bawp = 0;
 }
 
 static AVM_INLINE void init_interp_sf(INTERP_FILTER_SPEED_FEATURES *interp_sf) {
--- a/av2/encoder/speed_features.h
+++ b/av2/encoder/speed_features.h
@@ -846,6 +846,10 @@ typedef struct INTER_MODE_SPEED_FEATURES {
   // Prune the evaluation of ref_mv_idx > 0 for WARP_NEWMV modes based on RD
   // Cost of ref_mv_idx = 0 and best mode so far.
   bool prune_warp_newmv_ref_mv_idx;
+  // Histogram based frame level on/off decision for BAWP
+  // 0: no frame level on/off decision
+  // 1: enable histogram based frame level on/off
+  int enable_fast_bawp;
 } INTER_MODE_SPEED_FEATURES;
 
 typedef struct INTERP_FILTER_SPEED_FEATURES {


cmake/avm_config_defaults.cmake
```diff
@@ -149,6 +149,9 @@ set_avm_config_var(
   "Encoder only flag to configure encoder to enable mixed lossy/lossless coding"
 )
 
+set_avm_config_var(CONFIG_FAST_BAWP 0
```
yeqing-wu: For encoder optimization, we agreed not to use a compiler flag for the speed features. You can just add a new speed feature in the speed preset.

av2/av2_cx_iface.c
```diff
@@ -1535,6 +1535,10 @@ static avm_codec_err_t set_encoder_config(AV2EncoderConfig *oxcf,
 
   oxcf->speed = extra_cfg->cpu_used;
 
+#if CONFIG_FAST_BAWP
+  oxcf->enable_fast_bawp = oxcf->speed >= 1 ? 1 : 0;
```
yeqing-wu: It is better to add enable_fast_bawp as a new speed feature, instead of setting it in oxcf.

av2/encoder/encoder.h
```diff
@@ -1184,6 +1184,11 @@ typedef struct AV2EncoderConfig {
   // Indicates the spped preset to be used.
   int speed;
 
+#if CONFIG_FAST_BAWP
+  // Indicates whenter frame level on/off fast bawp is used.
+  int enable_fast_bawp;
```
yeqing-wu: Same comment: It is better to add enable_fast_bawp as a new speed feature, instead of setting it in oxcf.

av2/encoder/encodeframe.c
```diff
@@ -1952,6 +1952,73 @@ static AVM_INLINE void av2_enc_setup_tip_frame(AV2_COMP *cpi) {
...
+int av2_set_on_bawp_picture_level(AV2_COMP *cpi) {
+  int enable_curr_pic_bawp = 0;
+  assert(cpi->oxcf.enable_fast_bawp == 1);
+  AV2_COMMON *const cm = &cpi->common;
+  const YV12_BUFFER_CONFIG *source = cpi->source;
+
+  int numValues = 1 << cm->seq_params.bit_depth;
+
+  int32_t *curr_hist = (int32_t *)avm_calloc(numValues, sizeof(int32_t));
+  int32_t *ref_hist = (int32_t *)avm_calloc(numValues, sizeof(int32_t));
```
yeqing-wu: 1. Should we check whether the malloc succeeded, using CHECK_MEM_ERROR?
2. Also, even though this is a frame-level allocation/free, should we instead allocate and free it when the encoder is created and released? That would avoid the alloc/free on every frame and every pass.

av2/encoder/encodeframe.c
```diff
@@ -1952,6 +1952,73 @@ static AVM_INLINE void av2_enc_setup_tip_frame(AV2_COMP *cpi) {
   }
 }
 
+#if CONFIG_FAST_BAWP
+void get_histogram(int32_t *hist, const YV12_BUFFER_CONFIG *buf, int width,
```
yeqing-wu: Should we declare it as static if it's only used in this source file?

av2/encoder/encodeframe.c
```diff
@@ -1952,6 +1952,73 @@ static AVM_INLINE void av2_enc_setup_tip_frame(AV2_COMP *cpi) {
...
+  int stride = buf->y_stride;
+  for (int i = 0; i < height; i++) {
+    for (int j = 0; j < width; j++) {
+      hist[y_buf[j]]++;
+    }
+    y_buf += stride;
+  }
+}
+
+int av2_set_on_bawp_picture_level(AV2_COMP *cpi) {
```
yeqing-wu: Should we declare it as static if it's only used in this source file?

av2/encoder/encodeframe.c
```diff
@@ -1952,6 +1952,73 @@ static AVM_INLINE void av2_enc_setup_tip_frame(AV2_COMP *cpi) {
...
+        get_histogram(ref_hist, &ref_buf->buf, width, height);
+
+        // get SAD of delta histogram
+        int32_t diffHist = 0;
+        for (int i = 0; i < numValues; i++) {
+          diffHist += abs(curr_hist[i] - ref_hist[i]);
+        }
+
+        if (diffHist > hist_diff_thres) {
+          enable_curr_pic_bawp = 1;
```
yeqing-wu, line 1931: Once if (diffHist > hist_diff_thres) is satisfied, should we return here directly, to save the cost of the further histogram computation and checks? Based on the logic, in any case where diffHist > hist_diff_thres, enable_curr_pic_bawp would always be true, right? If yes, then we do not need enable_curr_pic_bawp

av2/encoder/encodeframe.c
```diff
@@ -1952,6 +1952,73 @@ static AVM_INLINE void av2_enc_setup_tip_frame(AV2_COMP *cpi) {
...
+
+  // get histogram of the current picture
+  get_histogram(curr_hist, source, width, height);
+
+  for (MV_REFERENCE_FRAME rf = 0; rf < cm->ref_frames_info.num_total_refs;
+       ++rf) {
+    const RefCntBuffer *ref_buf = get_ref_frame_buf(cm, rf);
+    if (ref_buf != NULL) {
+      int ref_width = ref_buf->mi_cols * 4;
+      int ref_height = ref_buf->mi_rows * 4;
```
yeqing-wu: For resolutions whose dimensions aren't multiples of 4, would the pixel values outside the frame (I can't recall whether the area outside the frame is padded or not) producing a wrong histogram and leading to a wrong decision?

av2/encoder/encodeframe.c
```diff
@@ -1952,6 +1952,73 @@ static AVM_INLINE void av2_enc_setup_tip_frame(AV2_COMP *cpi) {
...
+  const YV12_BUFFER_CONFIG *source = cpi->source;
+
+  int numValues = 1 << cm->seq_params.bit_depth;
+
+  int32_t *curr_hist = (int32_t *)avm_calloc(numValues, sizeof(int32_t));
+  int32_t *ref_hist = (int32_t *)avm_calloc(numValues, sizeof(int32_t));
+
+  int width = source->y_width;
+  int height = source->y_height;
+  int num_samples = width * height;
```
yeqing-wu: Use const if the variable's value won't be changed. Same comment for other variables inside this function.

av2/encoder/encodeframe.c
```diff
@@ -1952,6 +1952,73 @@ static AVM_INLINE void av2_enc_setup_tip_frame(AV2_COMP *cpi) {
...
+
+  for (MV_REFERENCE_FRAME rf = 0; rf < cm->ref_frames_info.num_total_refs;
+       ++rf) {
+    const RefCntBuffer *ref_buf = get_ref_frame_buf(cm, rf);
+    if (ref_buf != NULL) {
+      int ref_width = ref_buf->mi_cols * 4;
+      int ref_height = ref_buf->mi_rows * 4;
+
+      if (width == ref_width && height == ref_height) {
+        memset(ref_hist, 0, sizeof(int32_t) * numValues);
```
yeqing-wu: Let's use the AVM coding style for numValues and diffHist.

av2/encoder/encodeframe.c
yeqing-wu: Change it from `AV2_COMP *cpi ` to `const AV2_COMP *const cpi`, since we don't need to modify its members?

av2/encoder/encodeframe.c
jianle973: Yes, we should return immediately once (diffHist > hist_diff_thres) meets. In the code, this logic is fulfilled by 

        if (diff_hist > hist_diff_thres) {
          enable_curr_pic_bawp = 1;
          **break;**
        }

av2/encoder/encodeframe.c
jianle973: Yes, static is added.

av2/encoder/encodeframe.c
jianle973: Added static.

cmake/avm_config_defaults.cmake
jianle973: Done, thanks!

av2/av2_cx_iface.c
jianle973: Done, thanks!

av2/encoder/encoder.h
jianle973: Done, thanks!

av2/encoder/encodeframe.c
jianle973: 1) Added CHECK_MEM_ERROR. 
2) Since the memory size is small, I still keep it as frame level operation.

av2/encoder/encodeframe.c
jianle973: In my understanding, the pixels outside the input frames boundary are padded, and the reference also stored in the padded size. The padded areas also need to be encoded. It might be more accurate to also count them.

jianle973:
Thanks @yeqing-wu for reviewing the PR. I have tried to address all the comments in the new commit.

yeqing-wu (approved):
LGTM.

It is required to provide the anchor commit number and the A1/A2 results in the commit message. I think you can put this info in the PR conversation message. To prevent GitHub from messing up your table, you can wrap it with ``` before and after, like:

```
| Summary |   Y    |   U   |   V   |  YUV  | Enc-time | Dec-time |
| A1      | -0.02% | 0.24% | 0.21% | 0.00% | 97.8%    | 100.5%   |
| A2      | 0.10%  | 0.15% | 0.03% | 0.09% | 95.8%    | 99.8%    |
```

leolzhao:
Please add the separate class A1 and A2 results in the MR message? @jianle973

yunqingwang1 (approved):

jianle973:
33 frames results (speed 1) as follows:

| Summary |   Y    |   U   |   V   |  YUV  | Enc-time | Dec-time |
| A1      | +0.01% | +0.02% | +0.02% | +0.02% | 99.42%    | 100.70%   |
| A2     | +0.01%  | -0.02% | +0.13% | +0.02% | 99.34%    | 99.96%    |

yeqing-wu:
Since it doesn’t meet the acceptance criteria for speed 1, should we move it to speed 4? And what is the commit ID of the anchor?

leolzhao:
> Since it doesn’t meet the acceptance criteria for speed 1, should we move it to speed 4?

Yes, we can enable it for speed 4 if the resutls with cpu-used = 4 is similar or better.

jianle973:
> > Since it doesn’t meet the acceptance criteria for speed 1, should we move it to speed 4?
> 
> Yes, we can enable it for speed 4 if the resutls with cpu-used = 4 is similar or better.

It was 98.5% encoding time when I tested it on top v15. Not sure what's the problem. The running time deviation on our cluster could be the reason.

jianle973:
> > Since it doesn’t meet the acceptance criteria for speed 1, should we move it to speed 4?
> 
> Yes, we can enable it for speed 4 if the resutls with cpu-used = 4 is similar or better.

33 frame RA results for speed 2 and 3 are as follows, no loss with 0.5% encoding time reduction.
| Summary | Y | U | V | YUV | Enc-time | Dec-time |
Speed2, Overall w/o B2 (RA) +0.00% +0.01% +0.01% +0.00% 99.51% 100.12%
Speed3, Overall w/o B2 (RA) +0.00% +0.01% +0.03% +0.00% 99.45% 99.53%

Now, it's enabled for speed 2+.

@yeqing-wu @leolzhao

urvangjoshi:
@yeqing-wu @leolzhao : pls review again and resolve comments if ok.

yeqing-wu:
Based on the FG16 agreement, the encoding savings required for lossy changes should be at least 1% to prevent the encoding time savings from being negated by machine variations. However, if there are no losses, it is okay to me.

urvangjoshi:
@jianle973 : pls rebase and force-push to fix conflicts.

jianle973:
> @jianle973 : pls rebase and force-push to fix conflicts.

Rebased, thanks!

Issue #5142, opened 2026-07-22
Need to check the requirement of bitstream conformance that SeenFrameHeader is not equal to is_first_tile_group.


wantehchang:
This bug was discovered by libavif's avif_fuzztest_dec_incr_experimental and was originally reported in https://b.corp.google.com/issues/535185486.

### What version / commit were you testing with?

commit 636fcda84b06650b6b61493b4bf7337a9b102235

### What steps will reproduce the problem?

The AV2 specification, [Section 6.18 (Tile group OBU semantics) ](https://av2.aomedia.org/v1.0.0/index.html#tile_group_obu_semantics)has the following requirement of bitstream conformance:

> It is a requirement of bitstream conformance that SeenFrameHeader is not equal to is_first_tile_group.

The corresponding code in AVM is the `av2_read_tilegroup_header()` function in av2/decoder/decodeframe.c:

```
int32_t av2_read_tilegroup_header(
    AV2Decoder *pbi, struct avm_read_bit_buffer *rb, const uint8_t *data,
    const uint8_t **p_data_end, int *first_tile_group_in_frame, int *start_tile,
    int *end_tile, OBU_TYPE obu_type, int obu_xlayer_id) {
  ...
  int is_first_tile_group = 1;
  int send_uncompressed_header_flag = 1;
  bool send_first_tile_group_indication = true;
  ...
  if (send_first_tile_group_indication)
    is_first_tile_group = avm_rb_read_bit(rb);
  *first_tile_group_in_frame = is_first_tile_group;

  if (is_first_tile_group) {
    ...
  } else {
    send_uncompressed_header_flag = avm_rb_read_bit(rb);
  }

  if (send_uncompressed_header_flag) {
    if (!is_first_tile_group) {
      rb->bit_offset += pbi->uncomp_hdr_size_in_bits;
    } else {
      uint32_t uncomp_hdr_start_point = rb->bit_offset;
      read_uncompressed_header(pbi, obu_type, obu_xlayer_id, rb);

      ...
    }
  }
```

[...]

PR #5143, opened 2026-07-22
Check the requirement on is_first_tile_group

wantehchang:
Check the requirement of bitstream conformance that SeenFrameHeader is not equal to is_first_tile_group. If this requirement is not met, the decoder may skip the read_uncompressed_header() call in av2_read_tilegroup_header(), which eventually leads to a null pointer dereference.

Fixes https://github.com/AOMediaCodec/avm/issues/5142.

PR #5171, opened 2026-07-28
Check the requirement on is_first_tile_group (#5143)

wantehchang:
Check the requirement of bitstream conformance that SeenFrameHeader is not equal to is_first_tile_group. If this requirement is not met, the decoder may skip the read_uncompressed_header() call in av2_read_tilegroup_header(), which eventually leads to a null pointer dereference.

Fixes https://github.com/AOMediaCodec/avm/issues/5142.

(cherry picked from commit 63be7278689dcff55a60fa94e02feb4a67daaa8b)


wantehchang:
Leo: I suggest cherry-picking this bug fix to the av2-normative branch because it fixes a mismatch between the spec  and software.

PR #5172, opened 2026-07-28
Rework av2_update_nbr_diagonal_c

jingninghan:
Make the C and AVX2 code have the same behavior. Add unit test to check the two match.

### DIFF
--- a/av2/encoder/trellis_quant.c
+++ b/av2/encoder/trellis_quant.c
@@ -456,73 +456,76 @@ static int get_coeff_cost(int ci, tran_low_t abs_qc, int sign, int coeff_ctx,
 // Update neighbor coeff magnitudes state for current diagonal.
 void av2_update_nbr_diagonal_c(struct tcq_ctx_t *tcq_ctx, int row, int col,
                                int bwl) {
-  // Temp storage for next diagonal ctx, with padding.
-  uint8_t next_base_mag[32 + 8][TCQ_MAX_STATES];
-  uint8_t next_mid_mag[32 + 8][TCQ_MAX_STATES];
-  uint8_t(*next_base)[TCQ_MAX_STATES] = &next_base_mag[4];
-  uint8_t(*next_mid)[TCQ_MAX_STATES] = &next_mid_mag[4];
-  uint8_t(*mag_base)[TCQ_MAX_STATES] = &tcq_ctx->mag_base[4];
-  uint8_t(*mag_mid)[TCQ_MAX_STATES] = &tcq_ctx->mag_mid[4];
-
-  int idx_start = col;
-  int idx_end = 1 << bwl;
-  for (int st = 0; st < TCQ_MAX_STATES; st++) {
-    // Copy original coeff context from previous diagonal.
-    int orig_st = tcq_ctx->orig_st[st];
-    if (orig_st < 0) {
-      for (int i = 0; i < 32; i++) {
-        next_base[i][st] = 0;
-        next_mid[i][st] = 0;
-      }
-    } else {
-      for (int i = 0; i < 32; i++) {
-        next_base[i][st] = mag_base[i][orig_st];
-        next_mid[i][st] = mag_mid[i][orig_st];
-      }
-    }
-  }
-  memset(next_base_mag, 0, sizeof(next_base_mag[0]) * 4);
-  memset(next_mid_mag, 0, sizeof(next_mid_mag[0]) * 4);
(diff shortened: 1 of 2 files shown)


linzhen-beep (approved):

leolzhao:
Does it mean that the C code and AVX2 code for function av2_update_nbr_diagonal in av2-enc branch is mismatching before this commit? @jingninghan

jingninghan:
Yes, and i think the AVX2 version seems to make more sense, so went with its logic and fixed c version.

leolzhao (approved):

av2/encoder/trellis_quant.c
```diff
@@ -456,73 +456,77 @@ static int get_coeff_cost(int ci, tran_low_t abs_qc, int sign, int coeff_ctx,
...
   }
-  // Calc next context info
-  memset(tcq_ctx->ctx, 0, sizeof(tcq_ctx->ctx));
-  static const int8_t max_tbl[6] = { 0, 8, 6, 4, 4, 4 };
-  int base_max = max_tbl[AVMMIN(diag, 5)];
+
+  uint8_t new_mag_base[32 + 8][TCQ_MAX_STATES];
+  uint8_t new_mag_mid[32 + 8][TCQ_MAX_STATES];
+  memset(new_mag_base, 0, sizeof(new_mag_base));
+  memset(new_mag_mid, 0, sizeof(new_mag_mid));
```
yeqing-wu: Do we really need these two `memset` statements since every cell of `new_mag_base/new_mag_mid` is unconditionally assigned, not accumulated? This would waste unnecessary cycles (memset) to initialize these array in the hot path.

yeqing-wu: Please also check the memset for `state_arr` is also really necessary. To me, it is also unconditionally assigned, not accumulated

av2/encoder/trellis_quant.c
```diff
@@ -456,73 +456,77 @@ static int get_coeff_cost(int ci, tran_low_t abs_qc, int sign, int coeff_ctx,
...
+      int cur = state_arr[i][st];
+      state_arr[i + 1][st] = (cur < 0) ? -1 : tcq_ctx->prev_st[i][cur];
     }
   }
-  // Calc next context info
-  memset(tcq_ctx->ctx, 0, sizeof(tcq_ctx->ctx));
-  static const int8_t max_tbl[6] = { 0, 8, 6, 4, 4, 4 };
-  int base_max = max_tbl[AVMMIN(diag, 5)];
+
+  uint8_t new_mag_base[32 + 8][TCQ_MAX_STATES];
```
yeqing-wu: Should we change the magic number `32 + 8` to be `MAX_DIAG + 8`?

av2/encoder/trellis_quant.c
```diff
@@ -456,73 +456,77 @@ static int get_coeff_cost(int ci, tran_low_t abs_qc, int sign, int coeff_ctx,
...
-      // Update base positions {2, 0}, {1. 1}, {0, 2}
-      int base2 = AVMMIN(lev, max2);
-      mag_base[i][st] += base2;
-      mag_base[i - 1][st] += base2;
-      mag_base[i - 2][st] += base2;
-      // Update mid position {1, 1}
-      mag_mid[i - 1][st] = lev;
+    state_arr[idx0][st] = st;
+  }
+  for (int i = idx0; i < idx_end + 2; i++) {
```
yeqing-wu: Based on the compute loop code which reads the value of `state_arr[]`, it only reads up to `state_arr[idx_end + 1]`. However, in this for loop, it writes `state_arr[idx_end + 2]`. Therefore, should the end condition for the loop be `i < idx_end + 1`?

```
    for (int i = idx0; i < idx_end; i++) {
      ...
      int st0 = state_arr[i][st];
      int st1 = state_arr[i + 1][st];
      int st2 = state_arr[i + 2][st];
      ...
   }
```

av2/encoder/trellis_quant.c
jingninghan: removed

av2/encoder/trellis_quant.c
jingninghan: Done

av2/encoder/trellis_quant.c
jingninghan: Done

yeqing-wu (approved):
LGTM

PR #5174, opened 2026-07-29
Add the void cast to the tp_orig variable itself

jzern:
The void cast is used to ignore the tp_orig variable in Release builds
because tp_orig is only used in an assert(). So the void cast should be
applied to the variable itself rather than the expression *tp_orig.

This is a port from libaom:
 fd74e18c14 Add the void cast to the tp_orig variable itself


jzern:
cc: @wantehchang

wantehchang (approved):

PR #5175, opened 2026-07-29
extract_proto.cc: prefer std::size over ABSL_ARRAYSIZE

jzern:


jzern:
cc: @crmymh, @higher-performance

tools/extract_proto/extract_proto.cc
```diff
@@ -285,7 +286,7 @@ int PopulatePartitionTree(insp_frame_data *frame_data, insp_sb_data *sb_data,
 
   int child_coding_unit_start = coding_unit_start;
   int i = 0;
-  for (; i < (int)ABSL_ARRAYSIZE(tree->sub_tree); ++i) {
+  for (; i < (int)std::size(tree->sub_tree); ++i) {
```
wantehchang: I suggest declaring `i` as `size_t` and removing the `int` cast here.

wantehchang (approved):

tools/extract_proto/extract_proto.cc
jzern: Done.

wantehchang (approved):

PR #5177, opened 2026-07-30
Move lossless DPCM check inside prune_intra_y_mode

mudassir-sg:
This change moves condition checks related to lossless DPCM modes inside prune_intra_y_mode() and updates its return type from int to bool.

No stats changed.

### DIFF
--- a/av2/encoder/intra_mode_search.c
+++ b/av2/encoder/intra_mode_search.c
@@ -325,8 +325,14 @@ void av2_count_colors_highbd(const uint16_t *src, int stride, int rows,
   }
 }
 
-int prune_intra_y_mode(int64_t this_model_rd, int64_t *best_model_rd,
-                       int64_t top_intra_model_rd[], int k) {
+bool prune_intra_y_mode(int64_t this_model_rd, int64_t *best_model_rd,
+                        int64_t top_intra_model_rd[], int k, int lossless,
+                        uint8_t use_dpcm_y) {
+  (void)lossless;
+  assert(IMPLIES(use_dpcm_y != 0, lossless != 0));
+  if (this_model_rd < *best_model_rd) *best_model_rd = this_model_rd;
+  if (use_dpcm_y != 0) return false;
+
   const double thresh_top = 1.00;
   for (int i = 0; i < k; i++) {
     if (this_model_rd < top_intra_model_rd[i]) {
@@ -339,10 +345,9 @@ int prune_intra_y_mode(int64_t this_model_rd, int64_t *best_model_rd,
   }
   if (top_intra_model_rd[k - 1] != INT64_MAX &&
       this_model_rd > thresh_top * top_intra_model_rd[k - 1])
-    return 1;
+    return true;
 
-  if (this_model_rd < *best_model_rd) *best_model_rd = this_model_rd;
-  return 0;
+  return false;
 }
 
 #define PLANE_SIGN_TO_JOINT_SIGN(plane, a, b) \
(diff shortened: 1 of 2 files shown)


mudassir-sg:
@yunqingwang1 , @urvangjoshi 

av2/encoder/intra_mode_search.c
```diff
@@ -325,8 +325,12 @@ void av2_count_colors_highbd(const uint16_t *src, int stride, int rows,
   }
 }
 
-int prune_intra_y_mode(int64_t this_model_rd, int64_t *best_model_rd,
-                       int64_t top_intra_model_rd[], int k) {
+bool prune_intra_y_mode(const MACROBLOCKD *xd, int64_t this_model_rd,
+                        int64_t *best_model_rd, int64_t top_intra_model_rd[],
+                        int k) {
+  const MB_MODE_INFO *const mbmi = xd->mi[0];
+  if (xd->lossless[mbmi->segment_id] && mbmi->use_dpcm_y != 0) return false;
```
yunqingwang1: Can we only pass in (xd->lossless[mbmi->segment_id] && mbmi->use_dpcm_y != 0) to this function instead of xd?

mudassir-sg: Since `mbmi->use_dpcm_y=1` is applicable for lossless mode, checking `xd->lossless` is redundant. We have updated the function to take `mbmi->use_dpcm_y` directly.

yunqingwang1: "mbmi->use_dpcm_y=1 is applicable for lossless mode". Is there any code to guarantee this? Otherwise, could we at least add an assertion before calling prune_intra_y_mode()?

mudassir-sg: mbmi->use_dpcm_y = 1 is applicable only for lossless mode.
 
For example, in the Decoder (decodemv.c, line 1654), `mbmi->use_dpcm_y` is read from the bitstream only when `xd->lossless[mbmi->segment_id]` is set:
```
if (xd->lossless[mbmi->segment_id]) {
  mbmi->use_dpcm_y = read_dpcm_mode(ec_ctx, r);
```
Similar handling is present on the encoder side as well (in bitstream writing and intra mode search).
 
To make this invariant explicit, we have two options to add an assertion:

Option 1: Add assert(IMPLIES(mbmi->use_dpcm_y, lossless)) outside prune_intra_y_mode() at all 4 call sites.
Option 2: Pass lossless (or xd->lossless[mbmi->segment_id]) to prune_intra_y_mode() and add the assertion inside prune_intra_y_mode().
Please let us know which option would be preferred.

yunqingwang1: Either way is fine. Since one condition is being removed, let's add a check to ensure this logic isn't mistakenly broken in the future. Thanks.

yunqingwang1 (approved):

av2/encoder/intra_mode_search.c
mudassir-sg: Added the assertion inside `prune_intra_y_mode()` to guarantee that `use_dpcm_y` is set only for the lossless case.
 
During further testing, we noticed a bit-mismatch on the `Vertical_Bayshore_270x480_2997` sequence in the speed=0 Low-Delay (LD) configuration.
**Root Cause:** When evaluating lossless DPCM mode (`use_dpcm_y != 0`), the function returned early before updating `best_model_rd` and `top_intra_model_rd[]` . However, `best_model_rd` is later being used in:
- `model_intra_yrd_and_prune()` (for palette mode search)
- `rd_pick_intra_dip_sby()` (as an input feature to the ML model that prunes DIP modes)

(Note: While `top_intra_model_rd[]` is also bypassed during the early return for lossless DPCM, this is not an issue since `top_intra_model_rd[]` is reset prior to its usage).
**Fix:** Moved the `best_model_rd` update before the early exit check for `use_dpcm_y`.

yunqingwang1:
The failures happens at the current av2-enc branch. Will wait to rebase again after the revert PR gets in.

PR #5180, opened 2026-07-30
Enable deblock filter for partition research

jianle973:
This feature considers the distortion reduction of deblocking filter for partition research.

-- The deblocking filter is applied to a non-split partition after the best mode is picked for the current block. The distortion improvement of the deblocking filter is used to update the best RD cost of the current none-split partition block.

This encoder side only feature provides the coding gain with slight encoding time increase. Should be a good trade-off for low speed level.

Results of 33 frame (cpu1), a1 (missing 3 sequences) and a2 (missing 1 sequence) are as follows:

| Summary | Y | U | V | YUV | Enc-time | Dec-time |
| A1  | -0.62% |  -0.14% | -0.49% | -0.58% | 102.06% | 100.82% |
| A2 | -0.39% | +0.15% | -0.70% | -0.38% | 102.37% | 100.56% |

### DIFF
--- a/av2/encoder/encodeframe.c
+++ b/av2/encoder/encodeframe.c
@@ -2263,6 +2263,9 @@ static AVM_INLINE void encode_frame_internal(AV2_COMP *cpi) {
 
   av2_enc_setup_tip_frame(cpi);
 
+  cm->lf.apply_deblocking_filter[0] = cpi->oxcf.tool_cfg.enable_deblocking;
+  if (cm->lf.apply_deblocking_filter[0]) av2_loop_filter_frame_init(cm, 0, 1);
+
   cm->current_frame.skip_mode_info.skip_mode_flag =
       check_skip_mode_enabled(cpi) && cpi->oxcf.tool_cfg.enable_skip_mode;
 
--- a/av2/encoder/partition_search.c
+++ b/av2/encoder/partition_search.c
@@ -12,12 +12,14 @@
 
 #include "avm/avm_codec.h"
 #include "avm_ports/system_state.h"
+#include "avm_dsp/psnr.h"
 #include "av2/common/bru.h"
 #include "avm_ports/avm_timer.h"
 
 #include "av2/encoder/partition_mlp.h"
 
 #include "av2/common/av2_common_int.h"
+#include "av2/common/av2_loopfilter.h"
 #include "av2/common/blockd.h"
 #include "av2/common/common_data.h"
 #include "av2/common/enums.h"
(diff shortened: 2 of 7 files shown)


av2/encoder/partition_search.c
```diff
@@ -3562,10 +3565,90 @@ static void inter_sdp_copy_luma_mode_info(PC_TREE *pc_tree,
...
+    const int sub_y = cm->seq_params.subsampling_y;
+    rec_dist += avm_highbd_get_u_sse_part(
+        src, curr, mi_col * MI_SIZE >> sub_x, bw >> sub_x,
+        mi_row * MI_SIZE >> sub_y, bh >> sub_y);
+    rec_dist += avm_highbd_get_v_sse_part(
+        src, curr, mi_col * MI_SIZE >> sub_x, bw >> sub_x,
+        mi_row * MI_SIZE >> sub_y, bh >> sub_y);
+  }
+
+  if (labs(this_rdc->dist - rec_dist) < this_rdc->dist / 50) {
```
yeqing-wu: Both dist and rec_dist are of type int64_t, while labs() takes and returns long, which is 32-bit on LLP64 (MSVC/clang-cl x64) and all 32-bit targets. To avoid truncation, it is better to use llabs() with int64.

av2/encoder/encodeframe.c
```diff
@@ -2263,6 +2263,9 @@ static AVM_INLINE void encode_frame_internal(AV2_COMP *cpi) {
 
   av2_enc_setup_tip_frame(cpi);
 
+  cm->lf.apply_deblocking_filter[0] = cpi->oxcf.tool_cfg.enable_deblocking;
```
yeqing-wu, line 2266: Do we also need to initialize cm->lf.apply_deblocking_filter[1]?

av2/encoder/partition_search.c
```diff
@@ -3562,10 +3565,90 @@ static void inter_sdp_copy_luma_mode_info(PC_TREE *pc_tree,
...
+                                bsize);
+
+    av2_setup_dst_planes(pd, filtered, mi_row, mi_col, plane, plane + 1, NULL);
+    av2_filter_block_plane_horz(cm, xd, plane, &pd[plane], mi_row, mi_col, 1,
+                                bsize);
+
+    const int64_t filtered_y_dist = avm_highbd_get_y_sse_part(
+        src, filtered, h_start, area_w, v_start, area_h);
+
+    int sse_normal = 1 << ((10 - cm->seq_params.bit_depth) *
```
yeqing-wu: What if the bit depth is 12 bits? Is there any issue with that? How about handling different bit depths in this way?

```
const int shift = (10 - cm->seq_params.bit_depth) * 2;
const int64_t sse_normal = (shift >= 0) ? ((int64_t)1 << shift) : 0;  // >10-bit: skip the offset
if (!sse_normal) return 0;
```

av2/encoder/partition_search.c
```diff
@@ -3610,6 +3693,13 @@ static void none_partition_search(
   // Set PARTITION_NONE context and cost.
   set_none_partition_params(cm, td, x, pc_tree, part_search_state,
                             &best_remain_rdcost, best_rdc, &pt_cost);
+  if (best_rdc->rdcost != INT64_MAX &&
+      cpi->sf.lpf_sf.enable_deblock_for_partition_search) {
+    best_remain_rdcost.rate = best_remain_rdcost.rate * 9 / 8;
+    best_remain_rdcost.dist = best_remain_rdcost.dist * 9 / 8;
+    best_remain_rdcost.rdcost = best_remain_rdcost.rdcost * 9 / 8;
```
yeqing-wu, line 3707: Could we define macros for the numerator and denominator instead of using the magic numbers 9 and 8 here?

av2/encoder/partition_search.c
```diff
@@ -50,6 +50,9 @@
 #include "av2/encoder/partition_ml.h"
 #endif
 
+#include "avm_dsp/psnr.h"
+#include "av2/common/av2_loopfilter.h"
```
yeqing-wu: Move the two new #includes to alphabetical position and into the right category

yeqing-wu (commented):
This is a nice coding gain improvement. I’ve added some comments for your review. In the meantime, could you please update the commit or conversation messages to explain the changes in the code, so people can better understand them?

av2/encoder/encodeframe.c
jianle973: The filter is applied only to Luma component at the partition search stage in this implementation.

av2/encoder/partition_search.c
jianle973: The distortion in AVM is normalized to level of 16 * 8_bit_sse for cost calculation. I have revised the code to handle 12 bits.

av2/encoder/partition_search.c
jianle973: Good catch, thanks.

av2/encoder/partition_search.c
jianle973: Done.

av2/encoder/partition_search.c
jianle973: This is an experimental number. Explanation is provided in the updated version.

av2/encoder/speed_features.c
```diff
@@ -333,6 +333,8 @@ static void set_good_speed_features_framesize_independent(
   sf->mv_sf.fast_motion_estimation_on_block_256 = 1;
 
   sf->rd_sf.perform_coeff_opt = 1;
+
+  sf->lpf_sf.enable_deblock_for_partition_search = 1;
```
leolzhao: So, this speed feature is also enabled for cpu-used = 0? If so, do you have the results on top of cpu-used = 0?

jianle973:
> This is a nice coding gain improvement. I’ve added some comments for your review. In the meantime, could you please update the commit or conversation messages to explain the changes in the code, so people can better understand them?

Thanks @yeqing-wu for careful reviewing. I have made the change based on your comments. The completed a1/a2 results are also provided.

leolzhao:
One curious question: Why deblocking is only applied to none-split partition to improve the distortion instead of being applied to all the block partition types? @jianle973

jianle973:
> One curious question: Why deblocking is only applied to none-split partition to improve the distortion instead of being applied to all the block partition types? @jianle973

The words might be a bit confusing. The filter is applied to all the leaf coding blocks (e.g. none-split blocks) to recalculate the distortion. The RD_Cost of a splittable partition block accumulates the cost of its all leaf coding blocks, which means the RD_Cost updated by deblocking filter is taken into account.

av2/encoder/partition_search.c
```diff
@@ -3562,10 +3564,95 @@ static void inter_sdp_copy_luma_mode_info(PC_TREE *pc_tree,
...
+    AV2_COMP *const cpi, ThreadData *td, TileDataEnc *tile_data, MACROBLOCK *x,
+    TokenExtra **tp, PICK_MODE_CONTEXT *ctx_none, RD_STATS *const this_rdc,
+    const int mi_col, const int mi_row, const BLOCK_SIZE bsize) {
+  AV2_COMMON *const cm = &cpi->common;
+
+  const YV12_BUFFER_CONFIG *curr = &cm->cur_frame->buf;
+  const YV12_BUFFER_CONFIG *src = cpi->source;
+
+  const YV12_BUFFER_CONFIG *filtered = &cpi->trial_frame_rst;
+  int rate = 0;
```
yeqing-wu, line 3577: Is trial_frame_rst safe for multi-threaded encoding? Since we’ve enabled tile coding, I’m not sure if placing trial_frame_rst in cpi would cause race conditions or other threading issues during filter operations. Could you please check? 

I forgot to submit this comment yesterday.

jianle973: Good catch. I think it should be safe. To my understanding, in the current AVM software, the in-loop filter encoding process starts after the normal encoding process of the whole picture finishes.

leolzhao:
> > One curious question: Why deblocking is only applied to none-split partition to improve the distortion instead of being applied to all the block partition types? @jianle973
> 
> The words might be a bit confusing. The filter is applied to all the leaf coding blocks (e.g. none-split blocks) to recalculate the distortion. The RD_Cost of a splittable partition block accumulates the cost of its all leaf coding blocks, which means the RD_Cost updated by deblocking filter is taken into account.

Thank you for your explanation, and it makes sense to me.

yeqing-wu (approved):
LGTM except the Leo's question, if we want to enable it for speed 0, it is better to provide the results.

av2/encoder/speed_features.c
jianle973: OK, let me disable this feature for speed for this commit. I will test the speed 0 and submit another PR to enable it if the results are good.

jianle973:
> LGTM except the Leo's question, if we want to enable it for speed 0, it is better to provide the results.

Disable it for speed 0 now. Will test the performance later.

leolzhao (approved):

PR #5183, opened 2026-07-30
Add enable_low_complexity_decode (LC) mode

yunqingwang1:
Introduced an encoder option: "enable_low_complexity_decode" (LC mode) to target reduced decoding complexity. This mode is disabled by default. Also added related API control for ongoing LC mode development.

### DIFF
--- a/av2/av2_cx_iface.c
+++ b/av2/av2_cx_iface.c
@@ -1751,10 +1751,16 @@ static avm_codec_err_t set_encoder_config(AV2EncoderConfig *oxcf,
       extra_cfg->force_deferred_frames_for_ras_test;
 
   // Now, low complexity decode mode is only supported for good-quality
-  // encoding speed 1 and above. This can be further modified if needed.
+  // encoding. This can be further modified if needed.
   oxcf->enable_low_complexity_decode =
       extra_cfg->enable_low_complexity_decode &&
-      cfg->g_usage == AVM_USAGE_GOOD_QUALITY && oxcf->speed >= 1;
+      cfg->g_usage == AVM_USAGE_GOOD_QUALITY;
+  if (extra_cfg->enable_low_complexity_decode &&
+      cfg->g_usage != AVM_USAGE_GOOD_QUALITY) {
+    fprintf(stderr,
+            "Warning: setting enable_low_complexity_decode to 0 since it "
+            "requires good-quality usage.\n");
+  }
 
   if (update_config) {
     update_encoder_config(&cfg->encoder_cfg, extra_cfg);
--- a/av2/av2_cx_iface.c
+++ b/av2/av2_cx_iface.c
@@ -245,6 +245,7 @@ struct av2_extracfg {
   int buffer_refresh_multi_layers_test[REF_FRAMES];
   int multi_layers_lag_test;
   int force_deferred_frames_for_ras_test;
+  unsigned int enable_low_complexity_decode;
 };
 
 // Example subgop configs. Currently not used by default.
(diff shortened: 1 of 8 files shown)


yeqing-wu:
Hi @yunqingwang1 ,

LGTM, some minor comments for the documentation. Just one question: In this PR, we’ve only included the encoder command-line options for the LC decoder. However, we haven’t specified which features should be included in the enable_low_complexity_decode feature. It’s essentially a placeholder, right?

Thanks
Yeqing

av2/av2_cx_iface.c
```diff
@@ -1744,6 +1750,12 @@ static avm_codec_err_t set_encoder_config(AV2EncoderConfig *oxcf,
   oxcf->unit_test_cfg.force_deferred_frames_for_ras_test =
       extra_cfg->force_deferred_frames_for_ras_test;
 
+  // Now, low complexity decode mode is only supported for good-quality
+  // encoding speed 1 and above. This can be further modified if needed.
+  oxcf->enable_low_complexity_decode =
```
yeqing-wu, line 1755: If a user who runs `avmenc --enable-low-complexity-decode=1 --cpu-used=0` … gets the flag silently forced to 0 with no  warning or error. This is not good for a public-API contract. Should we emit a warning? Either reject or print a warning.

```
  if (extra_cfg->enable_low_complexity_decode &&
      (cfg->g_usage != AVM_USAGE_GOOD_QUALITY || oxcf->speed < 1)) {
    // Either reject:
    ERROR("enable_low_complexity_decode requires good-quality usage and speed>=1.");
    // Or at least fprintf a stderr warning before clearing the flag.
  }
```

yunqingwang1:
@yeqing-wu  Yes, it is a place holder, and features will be added later.

avm/avm_encoder.h
```diff
@@ -680,6 +680,11 @@ typedef struct cfg_options {
    *
    */
   int operating_points_count;
+
+  /*!\brief enable the low complexity decode mode
```
yeqing-wu, line 684: Since this would become the public API, should we include more details to document it?

- Document valid values (0 = disabled / 1 = enabled).
- State the preconditions the encoder actually enforces (good-quality usage, speed ≥ 1) — so callers can't
  tell why their flag "didn't take."
- Sentence must be capitalized for the first letter.

avm/avmcx.h
```diff
@@ -1249,6 +1249,11 @@ enum avme_enc_control_id {
    * the restricted_prediction_switch output ordering path.
    */
   AV2E_SET_FORCE_DEFERRED_FRAMES_FOR_RAS_TEST = 185,
+
+  /*!\brief Codec control function to enable the low complexity decode mode,
+   * unsigned int parameter. Value of zero means this mode is disabled.
+   */
```
yeqing-wu, line 1256: Also add the constraint to the comment:
the constraints (g_usage  == AVM_USAGE_GOOD_QUALITY, speed >= 1)

yeqing-wu (approved):

av2/av2_cx_iface.c
yunqingwang1: Thanks Yeqing. Added a warning. Also, I added the "speed >= 1" condition just for easy testing. In theory, we don't need it. I removed it.

avm/avm_encoder.h
yunqingwang1: Done.

avm/avmcx.h
yunqingwang1: Done.

yeqing-wu:
Thanks @yunqingwang1. The code looks good to me.

PR #5184, opened 2026-07-30
av2_get_chroma_format_idc: more monochrome checks

wantehchang:
In AV2, monochrome requires subsampling_x == 1 && subsampling_y == 1. (This somewhat arbitrary constraint comes from AV1.)

Also check the return value of av2_get_chroma_format_idc() at a few call sites.

### DIFF
--- a/av2/common/av2_common_int.h
+++ b/av2/common/av2_common_int.h
@@ -3075,7 +3075,11 @@ static INLINE avm_codec_err_t
 av2_get_chroma_format_idc(int subsampling_x, int subsampling_y, int monochrome,
                           uint32_t *seq_chroma_format_idc) {
   if (monochrome) {
-    *seq_chroma_format_idc = CHROMA_FORMAT_400;
+    if (subsampling_x == 1 && subsampling_y == 1) {
+      *seq_chroma_format_idc = CHROMA_FORMAT_400;
+    } else {
+      return AVM_CODEC_UNSUP_BITSTREAM;
+    }
   } else if (subsampling_x == 1 && subsampling_y == 1) {
     *seq_chroma_format_idc = CHROMA_FORMAT_420;
   } else if (subsampling_x == 1 && subsampling_y == 0) {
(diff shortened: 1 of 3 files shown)


av2/common/timing.c
```diff
@@ -60,9 +60,11 @@ int64_t av2_max_level_bitrate(BITSTREAM_PROFILE seq_profile, int seq_level_idx,
                               int subsampling_y, int monochrome) {
   int64_t bitrate;
 
-  uint32_t chroma_format_idc = CHROMA_FORMAT_420;
-  av2_get_chroma_format_idc(subsampling_x, subsampling_y, monochrome,
-                            &chroma_format_idc);
+  uint32_t chroma_format_idc;
+  if (av2_get_chroma_format_idc(subsampling_x, subsampling_y, monochrome,
+                                &chroma_format_idc) != AVM_CODEC_OK) {
+    return 0;
```
wantehchang: The (two) callers of this function treat a 0 return value as an error.

av2/common/av2_common_int.h
```diff
@@ -3075,7 +3075,11 @@ static INLINE avm_codec_err_t
 av2_get_chroma_format_idc(int subsampling_x, int subsampling_y, int monochrome,
                           uint32_t *seq_chroma_format_idc) {
   if (monochrome) {
-    *seq_chroma_format_idc = CHROMA_FORMAT_400;
+    if ((subsampling_x == 1 && subsampling_y == 1) {
```
jzern: Extra `(`

av2/common/timing.c
jzern: So it should be a bool return?

jzern (approved):

av2/common/av2_common_int.h
wantehchang: Done. Thanks!

av2/common/timing.c
wantehchang: No. This function returns a bitrate. But both callers of this function handle a 0 return value as an error. So if this `av2_get_chroma_format_idc()` call fails, we should return 0.

wantehchang: I found that this `av2_get_chroma_format_idc()` call is unneeded. So I reverted this change and will delete this `av2_get_chroma_format_idc()` call in PR https://github.com/AOMediaCodec/avm/pull/5185.

PR #5187, opened 2026-07-31
Check i<MAX_TILE_COLS in read_tile_syntax_info()

y-guyon:
Otherwise `width_sb` may not be 0.

---

https://av2.aomedia.org/v1.0.0/index.html#tile_params_semantics says:

> If uniform_tile_spacing_flag is equal to 0, it is a requirement of bitstream conformance that startSb is equal to sbCols when the loop writing sbColStarts exits.
>
> Note: The requirements on startSb ensure that the sizes of each tile add up to the full size of the frame when measured in superblocks.

So the `width_sb == 0` assertion at line 5959 makes sense.

However, the related code in https://av2.aomedia.org/v1.0.0/index.html#tile_params_syntax has no bound on `i`:

```py
widestTileSb = 1	
startSb = 0	
for ( i = 0; startSb < sbCols; i++ ) {	
  sbColStarts[ i ] = startSb	
  n = Min(sbCols - startSb, maxTileWidthSb)	
  width_in_sbs_minus_1	# ns(n)
  sizeSb = width_in_sbs_minus_1 + 1	
  widestTileSb = Max( sizeSb, widestTileSb )	
  startSb += sizeSb	
}
tileCols = i	
tileColsLog2 = tile_log2(1, tileCols)
```

whereas the implementation only loops as long as `i < MAX_TILE_COLS`.

---

This was found with a libavif fuzz target (b/540849406).

### DIFF
--- a/av2/decoder/decodeframe.c
+++ b/av2/decoder/decodeframe.c
@@ -4153,6 +4153,74 @@ static void reuse_tile_params(CommonTileParams *tiles,
   av2_calculate_tile_rows(tiles);
 }
 
+static void read_tile_syntax_info(CommonTileParams *tiles,
+                                  struct avm_read_bit_buffer *rb) {
+  int width_sb = tiles->sb_cols;
+  int height_sb = tiles->sb_rows;
+
+  tiles->uniform_spacing = avm_rb_read_bit(rb);
+
+  // Read tile columns
+  if (tiles->uniform_spacing) {
+    tiles->log2_cols = tiles->min_log2_cols;
+    while (tiles->log2_cols < tiles->max_log2_cols) {
+      if (!avm_rb_read_bit(rb)) {
+        break;
+      }
+      tiles->log2_cols++;
+    }
+  } else {
+    int i;
+    int start_sb;
+    for (i = 0, start_sb = 0; width_sb > 0; i++) {
+      if (i >= MAX_TILE_COLS) {
+        rb->error_handler(rb->error_handler_data, AVM_CODEC_CORRUPT_FRAME,
+                          "SeqTileCols cannot be greater than MAX_TILE_COLS");
+      }
+      const int size_sb =
+          1 + rb_read_uniform(rb, AVMMIN(width_sb, tiles->max_width_sb));
+      tiles->col_start_sb[i] = start_sb;
+      start_sb += size_sb;
+      width_sb -= size_sb;
+    }
+    assert(width_sb == 0);
+    assert(start_sb == tiles->sb_cols);
+    tiles->cols = i;
+    tiles->col_start_sb[i] = start_sb;
+  }
+  av2_calculate_tile_cols(tiles);
+
+  // Read tile rows
+  if (tiles->uniform_spacing) {
+    tiles->log2_rows = tiles->min_log2_rows;
+    while (tiles->log2_rows < tiles->max_log2_rows) {
+      if (!avm_rb_read_bit(rb)) {
+        break;
+      }
+      tiles->log2_rows++;
+    }
+  } else {
+    int i;
+    int start_sb;
+    for (i = 0, start_sb = 0; height_sb > 0; i++) {
+      if (i >= MAX_TILE_ROWS) {
+        rb->error_handler(rb->error_handler_data, AVM_CODEC_CORRUPT_FRAME,
+                          "SeqTileRows cannot be greater than MAX_TILE_ROWS");
+      }
+      const int size_sb =
+          1 + rb_read_uniform(rb, AVMMIN(height_sb, tiles->max_height_sb));
+      tiles->row_start_sb[i] = start_sb;
+      start_sb += size_sb;
+      height_sb -= size_sb;
+    }
+    assert(height_sb == 0);
+    assert(start_sb == tiles->sb_rows);
+    tiles->rows = i;
+    tiles->row_start_sb[i] = start_sb;
+  }
+  av2_calculate_tile_rows(tiles);
+}
+
 static AVM_INLINE void read_tile_info(AV2Decoder *const pbi,
                                       struct avm_read_bit_buffer *const rb) {
   AV2_COMMON *const cm = &pbi->common;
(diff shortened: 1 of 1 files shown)


wantehchang:
Yannis: Thank you for the fix. After the for loop exits, `tileRows` is set equal to `i`, and then `tileRows` is returned and stored in `TileRows`.  The bound on `i` can be found by searching for "TileRows` in the spec:

> **TileRows** specifies the number of tiles down the frame. It is a requirement of bitstream conformance that TileRows is less than or equal to MAX_TILE_ROWS.

av2/decoder/decodeframe.c
```diff
@@ -5947,7 +5947,14 @@ void read_tile_syntax_info(TileInfoSyntax *tile_params,
     int i;
     int start_sb;
     int width_sb = tile_info->sb_cols;
-    for (i = 0, start_sb = 0; width_sb > 0 && i < MAX_TILE_COLS; i++) {
+    for (i = 0, start_sb = 0; width_sb > 0; i++) {
+      if (i >= MAX_TILE_COLS) {
+        if (rb->error_handler) {
```
wantehchang: Omit the null pointer check for `rb->error_handler`. Also at line 5987.

Optional: You can add an assertion at the beginning of this function:

```
  // Verify rb has been configured to report errors.
  assert(rb->error_handler);
```

av2/decoder/decodeframe.c
```diff
@@ -5947,7 +5947,14 @@ void read_tile_syntax_info(TileInfoSyntax *tile_params,
...
+          rb->error_handler(rb->error_handler_data, AVM_CODEC_UNSUP_BITSTREAM,
+                            "The decoder only supports MAX_TILE_COLS tiles.");
```
wantehchang: Nit: Change "tiles" to "tile columns".

Similarly, at line 5989, change "tiles" to "tile rows".

av2/decoder/decodeframe.c
```diff
@@ -5947,7 +5947,14 @@ void read_tile_syntax_info(TileInfoSyntax *tile_params,
...
+        }
+        return;
```
wantehchang: Delete the `return` statement. Also at line 5991.

av2/decoder/decodeframe.c
wantehchang: I suggest moving the assertion immediately after the for loop, and use the fact that `width_sb` is 0:

```
    assert(width_sb == 0);
    tile_info->cols = i;
    tile_info->col_start_sb[i] = start_sb;
```

Optional: we can also assert `start_sb`:

```
    assert(start_sb == tile_info->sb_cols);
```

---

Make a similar change for rows below.

av2/decoder/decodeframe.c
y-guyon: Done.

av2/decoder/decodeframe.c
y-guyon: Since it is a question of bitstream conformance and not of decoder support, I went with `AVM_CODEC_CORRUPT_FRAME` and "TileCols cannot be greater than MAX_TILE_COLS".

av2/decoder/decodeframe.c
y-guyon: Done.

av2/decoder/decodeframe.c
y-guyon: Done.

av2/decoder/decodeframe.c
wantehchang: The new error code and error message make sense. Thanks.

Optional: I took a closer look and found that the corresponding variable in the spec should be `SeqTileCols` rather than `TileCols`. The spec does not specify a maximum value for `SeqTileCols`. But `SeqTileCols` is used to initialize `TileCols` when `reuse_tile_info` is equal to 1. This implies `SeqTileCols` should be subject to the same maximum value.

To be precise, we can consider changing `TileCols` to `SeqTileCols` in this error message. Similarly we can consider changing `TileRows` to `SeqTileRows` in this error message.

av2/decoder/decodeframe.c
```diff
@@ -5947,7 +5947,14 @@ void read_tile_syntax_info(TileInfoSyntax *tile_params,
     int i;
     int start_sb;
     int width_sb = tile_info->sb_cols;
-    for (i = 0, start_sb = 0; width_sb > 0 && i < MAX_TILE_COLS; i++) {
+    for (i = 0, start_sb = 0; width_sb > 0; i++) {
```
wantehchang: Optional: I will be happy to take care of this suggestion.

In the original code, if you search for "for (i = 0, start_sb = 0; width_sb > 0 && i < MAX_TILE_COLS; i++) {", you will see this pattern is used in two other functions:
- `reconstruct_tile_info_max_tile()`: After studying this function, I found that it can be simplified. See my PR https://github.com/AOMediaCodec/avm/pull/5190
- `read_tile_info_max_tile()`: I think we need to port your changes to this function. Please take a look.

wantehchang (approved):
LGTM. Thanks!

av2/decoder/decodeframe.c
y-guyon: > * `read_tile_info_max_tile()`: I think we need to port your changes to this function. Please take a look.

I went further and removed `read_tile_info_max_tile()` which is too similar to `read_tile_syntax_info()`.

av2/decoder/decodeframe.c
```diff
@@ -5947,19 +5888,22 @@ void read_tile_syntax_info(TileInfoSyntax *tile_params,
...
     }
-    tile_info->cols = i;
-    tile_info->col_start_sb[i] = start_sb + width_sb;
     assert(width_sb == 0);
+    assert(start_sb == tile_info->sb_cols);
+    tile_info->cols = i;
+    tile_info->col_start_sb[i] = start_sb;
   }
-  tile_info->min_log2_rows =
-      AVMMAX(tile_info->min_log2 - tile_info->log2_cols, 0);
```
y-guyon, line 5942: - `tile_info->min_log2_rows` is only read when `tile_info->uniform_spacing` is true.
- `av2_calculate_tile_cols()` always sets `tile_info->min_log2_rows` when `tile_info->uniform_spacing` is true.

Therefore this line can be safely removed. It was not part of `read_tile_info_max_tile()`.

av2/decoder/decodeframe.c
```diff
@@ -5928,10 +5871,8 @@ void av2_read_conformance_window(struct avm_read_bit_buffer *rb,
   }
 }
 
-void read_tile_syntax_info(TileInfoSyntax *tile_params,
+void read_tile_syntax_info(CommonTileParams *tile_info,
                            struct avm_read_bit_buffer *rb) {
-  tile_params->allow_tile_info_change = avm_rb_read_bit(rb);
```
y-guyon, line 5913: This line is moved to `read_sequence_tile_info()`. It was not part of `read_tile_info_max_tile()`.

av2/decoder/decodeframe.c
y-guyon: Done.

y-guyon:
Rebased to run the CI. Resolved conflicts with https://github.com/AOMediaCodec/avm/pull/5191.

av2/decoder/decodeframe.c
```diff
@@ -4193,15 +4192,21 @@ static AVM_INLINE void read_tile_info_max_tile(
   } else {
     int i;
     int start_sb;
-    for (i = 0, start_sb = 0; width_sb > 0 && i < MAX_TILE_COLS; i++) {
+    for (i = 0, start_sb = 0; width_sb > 0; i++) {
+      if (i >= MAX_TILE_COLS) {
+        rb->error_handler(rb->error_handler_data, AVM_CODEC_CORRUPT_FRAME,
+                          "SeqTileCols cannot be greater than MAX_TILE_COLS");
```
wantehchang: Note: Please make the same change to the error message at line 4228.

In the current code, this function has two callers. The first caller assigns this value to `SeqTileCols` while the second caller assigns this value to `TileCols`. So saying `SeqTileCols` in this error message is incorrect for the second caller.

One solution is to say "tileCols" (lowercase "t"), which is the name of the variable used in the `tile_params()` spec function. Another solution is to pass the prefix ("Seq" or "") as a function parameter.

av2/decoder/decodeframe.c
```diff
@@ -4173,9 +4173,8 @@ static void reconstruct_tile_info_max_tile(
   av2_calculate_tile_rows(tiles);
 }
 
-static AVM_INLINE void read_tile_info_max_tile(
-    AV2_COMMON *const cm, struct avm_read_bit_buffer *const rb) {
-  CommonTileParams *const tiles = &cm->tiles;
+static void read_tile_syntax_info(CommonTileParams *tiles,
+                                  struct avm_read_bit_buffer *rb) {
   int width_sb = tiles->sb_cols;
   int height_sb = tiles->sb_rows;
```
wantehchang: Optional: These two variables could be moved into the lexical scopes where they are used, as done in the original `read_tile_syntax_info()` function.

wantehchang (approved):
LGTM. I verified the changes are correct. Nice work!

av2/decoder/decodeframe.c
```diff
@@ -4173,9 +4173,8 @@ static void reconstruct_tile_info_max_tile(
   av2_calculate_tile_rows(tiles);
 }
 
-static AVM_INLINE void read_tile_info_max_tile(
-    AV2_COMMON *const cm, struct avm_read_bit_buffer *const rb) {
-  CommonTileParams *const tiles = &cm->tiles;
+static void read_tile_syntax_info(CommonTileParams *tiles,
```
wantehchang: Nit: I suggest renaming this function "read_tile_info" or "read_tile_params".

av2/decoder/decodeframe.c
wantehchang: Another reason this statement can be safely removed is that `tile_info->log2_cols` is not set unless `tile_info->uniform_spacing` is true.

Note that the `av2_calculate_tile_cols(tile_info)` call below will recalculate `tile_info->log2_cols`.

av2/decoder/decodeframe.c
wantehchang, line 4162: Optional: The use of `tiles->log2_cols` in this block can be replaced by a local variable, because the `av2_calculate_tile_cols(tiles)` call at line 4211 below will recalculate `tiles->log2_cols`.

wantehchang: Please ignore this suggestion. It is incorrect. Before `av2_calculate_tile_cols(tiles)` recalculates `tiles->log2_cols`, it already uses `tiles->log2_cols`.

av2/decoder/decodeframe.c
y-guyon: Done.

av2/decoder/decodeframe.c
y-guyon: `read_tile_info()` already exists. Done.

av2/decoder/decodeframe.c
y-guyon: Done.

wantehchang (approved):

PR #5188, opened 2026-07-31
Fix array subscript out of bounds warning in rdopt.c

jianj-g:
Use compact reference indices (COMPACT_INDEX0_NRS) when indexing ref_costs_comp in av2_rd_pick_inter_mode_sb to prevent out-of-bounds array access warning on 8x8 ref_costs_comp array.

### DIFF
--- a/av2/encoder/rdopt.c
+++ b/av2/encoder/rdopt.c
@@ -9274,9 +9274,10 @@ void av2_rd_pick_inter_mode_sb(struct AV2_COMP *cpi,
     }
 
     const int ref_frame_index = COMPACT_INDEX0_NRS(ref_frame);
-    const int ref_frame_cost = comp_pred
-                                   ? ref_costs_comp[ref_frame][second_ref_frame]
-                                   : ref_costs_single[ref_frame_index];
+    const int sec_ref_frame_index = COMPACT_INDEX1_NRS(second_ref_frame);
+    const int ref_frame_cost =
+        comp_pred ? ref_costs_comp[ref_frame_index][sec_ref_frame_index]
+                  : ref_costs_single[ref_frame_index];
     const int compmode_cost = (comp_ref_allowed && !is_tip_ref_frame(ref_frame))
                                   ? comp_inter_cost[comp_pred]
                                   : 0;


av2/encoder/rdopt.c
```diff
@@ -9274,9 +9274,10 @@ void av2_rd_pick_inter_mode_sb(struct AV2_COMP *cpi,
     }
 
     const int ref_frame_index = COMPACT_INDEX0_NRS(ref_frame);
-    const int ref_frame_cost = comp_pred
-                                   ? ref_costs_comp[ref_frame][second_ref_frame]
-                                   : ref_costs_single[ref_frame_index];
+    const int second_ref_frame_index = COMPACT_INDEX0_NRS(second_ref_frame);
```
wantehchang: I am not qualified to review this PR. git blame shows the original author might be Yeqing Wu or Hilmi Egilmez. It would be good to ask them to take a look.

av2/encoder/rdopt.c
```diff
@@ -9274,9 +9274,10 @@ void av2_rd_pick_inter_mode_sb(struct AV2_COMP *cpi,
...
+    const int ref_frame_cost =
+        comp_pred ? ref_costs_comp[ref_frame_index][second_ref_frame_index]
+                  : ref_costs_single[ref_frame_index];
```
wantehchang, line 9280: The `ref_costs_single` and `ref_costs_comp` arrays are declared as follows:

```
  unsigned int ref_costs_single[SINGLE_REF_FRAMES];
  ...
  unsigned int ref_costs_comp[MAX_COMPOUND_REF_INDEX][MAX_COMPOUND_REF_INDEX];
```

Note the `SINGLE_REF_FRAMES` vs. `MAX_COMPOUND_REF_INDEX`. So it is not clear why `ref_frame_index` can be used to index both arrays.

wantehchang (commented):
Jerome: I am not familiar with this code. Please

av2/encoder/rdopt.c
jianj-g: @yeqing-wu is the original author I think.

`MAX_COMPOUND_REF_INDEX` is 8 but `ref_frame` and `second_ref_frame` can be 36 (defined as `MV_REFERENCE_FRAME`) that's why compiler is complaining.
If `ref_frame` and `second_ref_frame` are guaranteed to be < 8 here then we should do something to silence the compiler warnings.

yeqing-wu: Is this warning only in av2-enc, and only after the inter-mode main loop refactor, right? If so, you can see that I use the ref_frame_centric_eval_order[] array, which clearly defines which compound combinations are possible. This is unlike the old code, where the second ref frame could be INTRA_FRAME (that is the reason why second_ref_frame could be 35 in the old code) and it relied on a condition check to decide whether that combination was allowed. With the new ref_frame_centric_eval_order[], the second ref frame is always within a reasonable range (<= 6).

yeqing-wu: I don't like the old code — there are so many rules to remember, and it's easy to make mistakes. So I made all the possible compound-mode ref frame combinations explicitly.

jianj-g: Yes, this warning only shows up in av2-enc.

Perhaps we can add an `assert` or something similar to silence the warnings. I'll see what I can do. Thanks for explanation!

urvangjoshi: FWIW, I'm also seeing this warnings.    I don't believe asserts will fix them in release build.
But @jianj-g can confirm.

jianj-g: The latest change silences the warnings.

av2/encoder/rdopt.c
```diff
@@ -9273,6 +9273,14 @@ void av2_rd_pick_inter_mode_sb(struct AV2_COMP *cpi,
...
     }
 
+    assert(IMPLIES(comp_pred, ref_frame < MAX_COMPOUND_REF_INDEX &&
+                                  second_ref_frame < MAX_COMPOUND_REF_INDEX));
+#if defined(__GNUC__)
+    if (comp_pred && (ref_frame >= MAX_COMPOUND_REF_INDEX ||
+                      second_ref_frame >= MAX_COMPOUND_REF_INDEX)) {
+      __builtin_unreachable();
+    }
+#endif
```
yeqing-wu: I’m not entirely sure why the warning is displayed. How about this solution to avoid the warning?

```
    const int ref_frame_index = COMPACT_INDEX0_NRS(ref_frame);
    const int sec_ref_frame_index = COMPACT_INDEX1_NRS(second_ref_frame);
    const int ref_frame_cost = comp_pred
                                   ? ref_costs_comp[ref_frame_index][sec_ref_frame_index]
                                   : ref_costs_single[ref_frame_index];
```

wantehchang: Yeqing: This is what Jerome proposed originally except that you use `COMPACT_INDEX1_NRS` instead of `COMPACT_INDEX0_NRS` with `second_ref_frame`.

When I reviewed an older version of this PR, I also wondered whether `COMPACT_INDEX1_NRS` (instead of `COMPACT_INDEX0_NRS`) should be used with `second_ref_frame`.

jianj-g: Yeqing, thanks for the suggestion. I wasn't aware of the differences between `COMPACT_INDEX1_NRS` and `COMPACT_INDEX0_NRS`.

yeqing-wu: @wantehchang, thanks for your explanation. If we genuinely need to add the assert and the compiler-related option to silence the warnings, then I believe @jianj-g’s initial proposal is more reasonable to me. Simply replace COMPACT_INDEX0_NRS with COMPACT_INDEX1_NRS for the second reference frame. Would that be your preference, @jianj-g?

yeqing-wu (approved):

wantehchang (approved):

PR #5190, opened 2026-07-31
Simplify and rename reconstruct_tile_info_max_tile

wantehchang:
reconstruct_tile_info_max_tile() is called after
read_tile_syntax_info(). So it can copy the tile_info->col_start_sb[] and tile_info->row_start_sb[] arrays and their sizes tile_info->cols and tile_info->rows directly.

Rename the function to reuse_tile_info().

### DIFF
--- a/av2/decoder/decodeframe.c
+++ b/av2/decoder/decodeframe.c
@@ -4124,51 +4124,31 @@ static AVM_INLINE void setup_frame_size_with_refs(
                                        &cm->error);
 }
 
-// Reconstructs the tile information
-static void reconstruct_tile_info_max_tile(
-    AV2_COMMON *const cm, const TileInfoSyntax *const tile_params) {
-  CommonTileParams *const tiles = &cm->tiles;
+// Reuses the tile information
+static void reuse_tile_params(CommonTileParams *tiles,
+                              const TileInfoSyntax *tile_params) {
   const CommonTileParams *const tile_info = &tile_params->tile_info;
-  int width_sb = tile_info->sb_cols;
-  int height_sb = tile_info->sb_rows;
 
   tiles->uniform_spacing = tile_info->uniform_spacing;
 
-  // Read tile columns
+  // Reuse tile columns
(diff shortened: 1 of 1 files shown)


av2/decoder/decodeframe.c
wantehchang: Peter: Please review. Thanks!

I hope I didn't miss some subtle reason as to why we must reconstruct tiles->col_start_sb[] and tiles->row_start_sb[]. I believe the original code is equivalent to simply copying from tile_info->col_start_sb[] and tile_info->row_start_sb[].

peterderivaz: I am away on holiday for a week so have not run any tests yet, but looking at the changes it certainly looks reasonable.

In the spec it looks like we have already done the same thing in reuse_tile_params (returning a copy of sequence level arrays instead of reconstructing tile by tile) so it is probably okay.

The thing that I remember you need to be careful about is when switching between intra frames and inter frames with superblock size of 256 by 256, because the intra frames are forced to a 128 by 128 superblock size, and sometimes there are interesting corner cases when the inter frames are exactly twice the size of the intra frames.

av2/decoder/decodeframe.c
```diff
@@ -4251,7 +4231,7 @@ static AVM_INLINE void read_tile_info(AV2Decoder *const pbi,
       reuse = 1;
   }
   if (reuse) {
-    reconstruct_tile_info_max_tile(cm, tile_params);
+    reuse_tile_info(&cm->tiles, tile_params);
```
wantehchang: Yannis: With your work on PR https://github.com/AOMediaCodec/avm/pull/5187, you have become one of the best persons to review this PR. Thanks!

y-guyon (approved):

wantehchang:
Urvang, Yannis: I renamed the function to `reuse_tile_params` to be consistent with the latest `read_tile_params` function name in PR https://github.com/AOMediaCodec/avm/pull/5187.

PR #5193, opened 2026-08-03
Prune uneven 4A/4B partitions with neighbor block boundaries

urvangjoshi:
Use the best partitioning boundaries of left / above coding blocks to prune uneven 4way partitions -- by checking boundary alignment.

Only prune for inter frames.

- Enabled for speed >= 1.
- But disabled for resolution <= 270p.

Results for RA CTC 33 frames, speed 1:
(Anchor: 8d9588497)

```
| Summary |   Y    |   U    |   V    |  YUV   | Enc-time |
| A1      |  0.08% |  0.04% |  0.11% |  0.08% | 97.10%   |
| A2      |  0.07% | -0.07% |  0.16% |  0.07% | 96.50%   |
|avg wo b2|  0.07% |  0.09% |  0.06% |  0.07% | 96.75%   |
```

Note: no changes at speed 4, because uneven 4way partitions are disabled at this speed.

Related: refactor a common av2_get_chroma_start_location() function.

STATS_CHANGED for speed >= 1.

### DIFF
--- a/av2/common/av2_common_int.h
+++ b/av2/common/av2_common_int.h
@@ -6050,6 +6050,21 @@ static INLINE bool av2_skip_reference_buffer_update(
   return clear_multiple_insert_in_one && ref_index != first_ref_index;
 }
 
+// Returns the starting mi location of chroma reference block for the current
+// mbmi, by setting `chroma_mi_row_start` and `chroma_mi_col_start`.
+static INLINE void av2_get_chroma_start_location(const MB_MODE_INFO *mbmi,
+                                                 TREE_TYPE tree_type,
+                                                 int *chroma_mi_row_start,
+                                                 int *chroma_mi_col_start) {
+  if (tree_type == SHARED_PART) {
+    *chroma_mi_row_start = mbmi->chroma_ref_info.mi_row_chroma_base;
+    *chroma_mi_col_start = mbmi->chroma_ref_info.mi_col_chroma_base;
+  } else {
+    *chroma_mi_row_start = mbmi->chroma_mi_row_start;
+    *chroma_mi_col_start = mbmi->chroma_mi_col_start;
+  }
+}
+
 #ifdef __cplusplus
 }  // extern "C"
 #endif
--- a/av2/common/av2_common_int.h
+++ b/av2/common/av2_common_int.h
@@ -6056,6 +6056,7 @@ static INLINE void av2_get_chroma_start_location(const MB_MODE_INFO *mbmi,
                                                  TREE_TYPE tree_type,
                                                  int *chroma_mi_row_start,
                                                  int *chroma_mi_col_start) {
+  assert(tree_type == SHARED_PART || tree_type == CHROMA_PART);
   if (tree_type == SHARED_PART) {
     *chroma_mi_row_start = mbmi->chroma_ref_info.mi_row_chroma_base;
     *chroma_mi_col_start = mbmi->chroma_ref_info.mi_col_chroma_base;
--- a/av2/encoder/partition_search.c
+++ b/av2/encoder/partition_search.c
@@ -2925,19 +2925,13 @@ static AVM_INLINE bool is_same_block_for_tree(const MB_MODE_INFO *m1,
 static AVM_INLINE void prune_4way_partitions_with_neighbor_boundaries(
     PartitionSearchState *state, const AV2_COMMON *cm, const MACROBLOCKD *xd,
     int mi_row, int mi_col, BLOCK_SIZE bsize) {
-  const int mi_width = mi_size_wide[bsize];
+  // Check left neighbor for horizontal boundaries using mi array
   const int mi_height = mi_size_high[bsize];
-
   const int available_mi_height =
       AVMMIN(mi_height, cm->mi_params.mi_rows - mi_row);
-  const int available_mi_width =
-      AVMMIN(mi_width, cm->mi_params.mi_cols - mi_col);
-
-  bool left_horz_boundaries[MAX_MIB_SIZE] = { false };
-  bool top_vert_boundaries[MAX_MIB_SIZE] = { false };
-
-  // Check left neighbor for horizontal boundaries using mi array
-  if (xd->left_available && mi_height >= 8) {
+  if (xd->left_available && mi_height >= 8 &&
+      available_mi_height == mi_height) {
+    bool left_horz_boundaries[MAX_MIB_SIZE] = { false };
     for (int r = 1; r < available_mi_height; r++) {
       const MB_MODE_INFO *m1 = xd->mi[r * xd->mi_stride - 1];
       const MB_MODE_INFO *m2 = xd->mi[(r - 1) * xd->mi_stride - 1];
@@ -2945,21 +2939,7 @@ static AVM_INLINE void prune_4way_partitions_with_neighbor_boundaries(
         left_horz_boundaries[r] = true;
       }
     }
-  }
-
-  // Check top neighbor for vertical boundaries using mi array
-  if (xd->up_available && mi_width >= 8) {
-    for (int c = 1; c < available_mi_width; c++) {
-      const MB_MODE_INFO *m1 = xd->mi[-xd->mi_stride + c];
-      const MB_MODE_INFO *m2 = xd->mi[-xd->mi_stride + c - 1];
-      if (!is_same_block_for_tree(m1, m2, xd->tree_type)) {
-        top_vert_boundaries[c] = true;
-      }
-    }
-  }
-
-  // Prune 4-way Partitions.
-  if (xd->left_available && mi_height >= 8) {
+    // Prune HORZ 4A/4B partitions.
     if (!left_horz_boundaries[mi_height / 8] &&
         !left_horz_boundaries[3 * mi_height / 8] &&
         !left_horz_boundaries[7 * mi_height / 8]) {
@@ -2972,7 +2952,20 @@ static AVM_INLINE void prune_4way_partitions_with_neighbor_boundaries(
     }
   }
 
-  if (xd->up_available && mi_width >= 8) {
+  // Check top neighbor for vertical boundaries using mi array
+  const int mi_width = mi_size_wide[bsize];
+  const int available_mi_width =
+      AVMMIN(mi_width, cm->mi_params.mi_cols - mi_col);
+  if (xd->up_available && mi_width >= 8 && available_mi_width == mi_width) {
+    bool top_vert_boundaries[MAX_MIB_SIZE] = { false };
+    for (int c = 1; c < available_mi_width; c++) {
+      const MB_MODE_INFO *m1 = xd->mi[-xd->mi_stride + c];
+      const MB_MODE_INFO *m2 = xd->mi[-xd->mi_stride + c - 1];
+      if (!is_same_block_for_tree(m1, m2, xd->tree_type)) {
+        top_vert_boundaries[c] = true;
+      }
+    }
+    // Prune VERT 4A/4B partitions.
     if (!top_vert_boundaries[mi_width / 8] &&
         !top_vert_boundaries[3 * mi_width / 8] &&
         !top_vert_boundaries[7 * mi_width / 8]) {
--- a/av2/encoder/partition_search.c
+++ b/av2/encoder/partition_search.c
@@ -2900,6 +2900,83 @@ static AVM_INLINE void init_allowed_partitions(
 #endif  // CONFIG_COLLECT_PARTITION_STATS
 }
 
+static AVM_INLINE bool is_same_block_for_tree(const MB_MODE_INFO *m1,
+                                              const MB_MODE_INFO *m2,
+                                              TREE_TYPE tree_type) {
+  if (!m1 || !m2) return false;
+
+  if (tree_type == CHROMA_PART) {
+    int m1_r;
+    int m1_c;
+    av2_get_chroma_start_location(m1, m1->tree_type, &m1_r, &m1_c);
+    int m2_r;
+    int m2_c;
+    av2_get_chroma_start_location(m2, m2->tree_type, &m2_r, &m2_c);
+    return m1_r == m2_r && m1_c == m2_c;
+  } else {
+    return m1->mi_row_start == m2->mi_row_start &&
+           m1->mi_col_start == m2->mi_col_start;
+  }
+}
+
+static AVM_INLINE void prune_partitions_with_neighbor_boundaries(
+    PartitionSearchState *state, const AV2_COMMON *cm, const MACROBLOCKD *xd,
+    int mi_row, int mi_col, BLOCK_SIZE bsize) {
+  const int mi_width = mi_size_wide[bsize];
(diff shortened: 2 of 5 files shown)


leolzhao:
The trade off for class A2 is pretty good, however the trade off for class A1 is worse than the threshold. Currently, our rule is to make sure the trade off for both class A1 and A2 should satisfy the threshold. So, is it possible to add the qp dependent condition for this speed feature? For example, only turn on this speed feature for 4K sequence when qp is greater than one threshold? @urvangjoshi

urvangjoshi:
> The trade off for class A2 is pretty good, however the trade off for class A1 is worse than the threshold. Currently, our rule is to make sure the trade off for both class A1 and A2 should satisfy the threshold. So, is it possible to add the qp dependent condition for this speed feature? For example, only turn on this speed feature for 4K sequence when qp is greater than one threshold? @urvangjoshi

Thanks @leolzhao for the suggestion!  Let me try some variations in this direction.

av2/common/av2_common_int.h
```diff
@@ -6050,6 +6050,21 @@ static INLINE bool av2_skip_reference_buffer_update(
   return clear_multiple_insert_in_one && ref_index != first_ref_index;
 }
 
+// Returns the starting mi location of chroma reference block for the current
+// mbmi, by setting `chroma_mi_row_start` and `chroma_mi_col_start`.
+static INLINE void av2_get_chroma_start_location(const MB_MODE_INFO *mbmi,
+                                                 TREE_TYPE tree_type,
+                                                 int *chroma_mi_row_start,
+                                                 int *chroma_mi_col_start) {
```
yeqing-wu, line 6058: Any reason to remove the original assert 'assert(tree_type == SHARED_PART || tree_type == CHROMA_PART);'?

av2/encoder/partition_search.c
```diff
@@ -2896,6 +2896,83 @@ static AVM_INLINE void init_allowed_partitions(
...
+  if (!m1 || !m2) return false;
+
+  if (tree_type == CHROMA_PART) {
+    int m1_r;
+    int m1_c;
+    av2_get_chroma_start_location(m1, m1->tree_type, &m1_r, &m1_c);
+    int m2_r;
+    int m2_c;
+    av2_get_chroma_start_location(m2, m2->tree_type, &m2_r, &m2_c);
+    return m1_r == m2_r && m1_c == m2_c;
```
yeqing-wu, line 2919: If the neighbors m1 and m2 tree type is LUMA_PART, what would happen? Or could this happen?

av2/encoder/partition_search.c
```diff
@@ -2896,6 +2896,83 @@ static AVM_INLINE void init_allowed_partitions(
...
+  } else {
+    return m1->mi_row_start == m2->mi_row_start &&
+           m1->mi_col_start == m2->mi_col_start;
+  }
+}
+
+static AVM_INLINE void prune_partitions_with_neighbor_boundaries(
```
leolzhao: Please add a comment describing the function. Mention that it prunes uneven 4-way partitions by checking if neighboring top/left MI units share the same block ID, avoiding potential confusion with boundary detection algorithms.

av2/encoder/speed_features.c
```diff
@@ -1102,6 +1104,7 @@ static AVM_INLINE void set_erp_speed_features(AV2_COMP *cpi) {
     // Emulate erp_pruning_level = 6.
     sf->part_sf.ext_recur_depth_level = 1;
     sf->part_sf.ml_early_term_after_part_split_level = 2;
+    sf->part_sf.prune_part_with_neighbor_boundaries = true;
```
yeqing-wu: We always enable sf->part_sf.prune_part_with_neighbor_boundaries for speed >= 1, and also set its value to be true for erp_pruning_level is 6. What should be the behavior if erp_pruning_level is not 6? If still need to turn on prune_part_with_neighbor_boundaries, then the set of prune_part_with_neighbor_boundaries under erp_pruning_level is 6 is redundant, right?

av2/encoder/partition_search.c
```diff
@@ -2896,6 +2896,83 @@ static AVM_INLINE void init_allowed_partitions(
...
+  const int available_mi_height =
+      AVMMIN(mi_height, cm->mi_params.mi_rows - mi_row);
+  const int available_mi_width =
+      AVMMIN(mi_width, cm->mi_params.mi_cols - mi_col);
+
+  bool left_horz_boundaries[MAX_MIB_SIZE] = { false };
+  bool top_vert_boundaries[MAX_MIB_SIZE] = { false };
+
+  // Check left neighbor for horizontal boundaries using mi array
+  if (xd->left_available && xd->mi_col > 0) {
```
yeqing-wu: Is this xd->mi_col > 0 redundant since if xd->left_available is valid, then mi_col is always > 0, right?

av2/encoder/partition_search.c
```diff
@@ -2896,6 +2896,83 @@ static AVM_INLINE void init_allowed_partitions(
...
+      const MB_MODE_INFO *m1 = xd->mi[r * xd->mi_stride - 1];
+      const MB_MODE_INFO *m2 = xd->mi[(r - 1) * xd->mi_stride - 1];
+      if (!is_same_block_for_tree(m1, m2, xd->tree_type)) {
+        left_horz_boundaries[r] = true;
+      }
+    }
+  }
+
+  // Check top neighbor for vertical boundaries using mi array
+  if (xd->up_available && xd->mi_row > 0) {
```
yeqing-wu: same for xd->mi_row > 0

av2/encoder/partition_search.c
```diff
@@ -2896,6 +2896,83 @@ static AVM_INLINE void init_allowed_partitions(
...
+      const MB_MODE_INFO *m1 = xd->mi[-xd->mi_stride + c];
+      const MB_MODE_INFO *m2 = xd->mi[-xd->mi_stride + c - 1];
+      if (!is_same_block_for_tree(m1, m2, xd->tree_type)) {
+        top_vert_boundaries[c] = true;
+      }
+    }
+  }
+
+  // Prune 4-way Partitions.
+  if (xd->left_available && xd->mi_col > 0) {
```
yeqing-wu: same comment

av2/encoder/partition_search.c
```diff
@@ -2896,6 +2896,83 @@ static AVM_INLINE void init_allowed_partitions(
...
+    if (mi_height >= 8 && (!left_horz_boundaries[mi_height / 8] &&
+                           !left_horz_boundaries[3 * mi_height / 8] &&
+                           !left_horz_boundaries[5 * mi_height / 8] &&
+                           !left_horz_boundaries[7 * mi_height / 8])) {
+      state->prune_partition[PARTITION_HORZ_4A] = true;
+      state->prune_partition[PARTITION_HORZ_4B] = true;
+    }
+  }
+
+  if (xd->up_available && xd->mi_row > 0) {
```
yeqing-wu: same comment

av2/encoder/partition_search.c
```diff
@@ -2896,6 +2896,83 @@ static AVM_INLINE void init_allowed_partitions(
...
+      }
+    }
+  }
+
+  // Prune 4-way Partitions.
+  if (xd->left_available && xd->mi_col > 0) {
+    if (mi_height >= 8 && (!left_horz_boundaries[mi_height / 8] &&
+                           !left_horz_boundaries[3 * mi_height / 8] &&
+                           !left_horz_boundaries[5 * mi_height / 8] &&
+                           !left_horz_boundaries[7 * mi_height / 8])) {
```
yeqing-wu: We only read 4 slots here, however we allocate 32 slots, should we try to optimize it?

`  bool left_horz_boundaries[MAX_MIB_SIZE] = { false };`

av2/encoder/partition_search.c
```diff
@@ -2896,6 +2896,83 @@ static AVM_INLINE void init_allowed_partitions(
...
+      state->prune_partition[PARTITION_HORZ_4A] = true;
+      state->prune_partition[PARTITION_HORZ_4B] = true;
+    }
+  }
+
+  if (xd->up_available && xd->mi_row > 0) {
+    if (mi_width >= 8 && (!top_vert_boundaries[mi_width / 8] &&
+                          !top_vert_boundaries[3 * mi_width / 8] &&
+                          !top_vert_boundaries[5 * mi_width / 8] &&
+                          !top_vert_boundaries[7 * mi_width / 8])) {
```
yeqing-wu: same comment

av2/encoder/partition_search.c
```diff
@@ -2896,6 +2896,83 @@ static AVM_INLINE void init_allowed_partitions(
...
+  const int mi_width = mi_size_wide[bsize];
+  const int mi_height = mi_size_high[bsize];
+
+  const int available_mi_height =
+      AVMMIN(mi_height, cm->mi_params.mi_rows - mi_row);
+  const int available_mi_width =
+      AVMMIN(mi_width, cm->mi_params.mi_cols - mi_col);
+
+  bool left_horz_boundaries[MAX_MIB_SIZE] = { false };
+  bool top_vert_boundaries[MAX_MIB_SIZE] = { false };
```
yeqing-wu: if available_mi_height < mi_height, then [1, 3, 5, 7] only 1 would be available, 3, 5, 7 would be false, should we return it immediately?

```
    if (mi_height >= 8 && (!left_horz_boundaries[mi_height / 8] &&
                           !left_horz_boundaries[3 * mi_height / 8] &&
                           !left_horz_boundaries[5 * mi_height / 8] &&
                           !left_horz_boundaries[7 * mi_height / 8])) {
```

yeqing-wu: Same comment for available_mi_width < mi_width

`if (available_mi_height < mi_height || available_mi_width < mi_width) return;`

av2/encoder/partition_search.c
```diff
@@ -2896,6 +2896,83 @@ static AVM_INLINE void init_allowed_partitions(
...
+
+  // Check left neighbor for horizontal boundaries using mi array
+  if (xd->left_available && xd->mi_col > 0) {
+    for (int r = 1; r < available_mi_height; r++) {
+      const MB_MODE_INFO *m1 = xd->mi[r * xd->mi_stride - 1];
+      const MB_MODE_INFO *m2 = xd->mi[(r - 1) * xd->mi_stride - 1];
```
yeqing-wu, line 2941: what would happen if `(r - 1) * xd->mi_stride - 1` cross the tile boundary?

av2/encoder/partition_search.c
```diff
@@ -2896,6 +2896,83 @@ static AVM_INLINE void init_allowed_partitions(
...
+        left_horz_boundaries[r] = true;
+      }
+    }
+  }
+
+  // Check top neighbor for vertical boundaries using mi array
+  if (xd->up_available && xd->mi_row > 0) {
+    for (int c = 1; c < available_mi_width; c++) {
+      const MB_MODE_INFO *m1 = xd->mi[-xd->mi_stride + c];
+      const MB_MODE_INFO *m2 = xd->mi[-xd->mi_stride + c - 1];
```
yeqing-wu, line 2964: same here, what if `-xd->mi_stride + c - 1` cross the tile boundary from the left tile?

av2/encoder/speed_features.h
```diff
@@ -441,6 +441,8 @@ typedef struct PARTITION_SPEED_FEATURES {
   //    intra coded block, prunes when inter ratio exceeds 50%, and early skips
   //    when current best partitioning is PARTITION_NONE.
   int inter_sdp_fast_method_level;
+  // Prune partition types if they don't align with neighbor block boundaries.
+  bool prune_part_with_neighbor_boundaries;
```
yeqing-wu: all other items in this data structure is 'int', should we keep them be consistent?

yeqing-wu:
Github uses markdown style in its conversation messages. To make sure that the results are not messed up by Github, please add ``` before and after the results.

av2/common/av2_common_int.h
urvangjoshi: Good catch! Added it back.

av2/encoder/partition_search.c
urvangjoshi: If tree type is luma, it will go into the else condition.  We can't have a condition where `xd->tree_type` is chroma, but m1/m2->tree_type is luma.

av2/encoder/partition_search.c
urvangjoshi: We are starting from `r = 1`  and also `left_available == true` here,  so this will never cross tile boundary.  Am I missing something?

av2/encoder/partition_search.c
urvangjoshi: same here

av2/encoder/speed_features.h
urvangjoshi: Done

av2/encoder/speed_features.c
urvangjoshi: Currently, the erp_pruning_level and speed are set independently from command-line.   So, we need to ensure both options set speed features appropriately.
As seen in the comment a few lines above this line,  we need to duplicate the speed feature setting here for this reason.
Hope that answers your question.

av2/encoder/partition_search.c
urvangjoshi: Indeed, removed!

av2/encoder/partition_search.c
urvangjoshi: Indeed, removed!

av2/encoder/partition_search.c
urvangjoshi: see other reply

av2/encoder/partition_search.c
urvangjoshi: Great point!  I have rewritten the code to add this check,  and in general simplify the code to avoid repeated checks.

urvangjoshi:
> The trade off for class A2 is pretty good, however the trade off for class A1 is worse than the threshold. Currently, our rule is to make sure the trade off for both class A1 and A2 should satisfy the threshold. So, is it possible to add the qp dependent condition for this speed feature? For example, only turn on this speed feature for 4K sequence when qp is greater than one threshold? @urvangjoshi

@leolzhao :  I made some tweaks to the code as follows:
- Rewrote the code to remove some checks and add some higher level checks to avoid some computations.
- Disabled this speed feature for <= 270p resolution,  as I was seeing a large loss on a5 set only (~0.3%).
- Reran on a cluster with more reliable timing.

With that, all sets seem to have a good trade-off now.  See the updated commit message for latest results.

av2/encoder/partition_search.c
yeqing-wu: I think you are right, I missed the r is starting from 1. Sorry for this.

leolzhao (approved):

yeqing-wu (approved):

urvangjoshi:
@leolzhao :  I'm not sure why the CI run is failing in "common variable setup" phase.  Seems to be a transient issue perhaps.
You may need to force-merge the PR.

PR #5204, opened 2026-08-04
Add NEON for highbd convolve8, convolve_copy, and upsampled_pred

jjustiss-apple:
This adds NEON intrinsics for convolve8_horiz (11x speedup),
convolve8_vert (13x speedup), convolve_copy, and the upsampled_pred
dispatch that enables the convolve8 NEON path to be reached from the
motion estimation variance computation.

CTC results (RA, cpu-used=1, 33 frames, A5):

| Metric | Delta |
|--------|-------|
| Encode time | -11.02% |
| BD-rate Y | 0.000% |
| BD-rate Cb | 0.000% |
| BD-rate Cr | 0.000% |

Bit-exact confirmed (all PSNR values identical).

Unit tests included.


avm_dsp/avm_convolve.c
```diff
@@ -161,7 +153,7 @@ void avm_highbd_convolve_copy_c(const uint16_t *src, ptrdiff_t src_stride,
                                 uint16_t *dst, ptrdiff_t dst_stride, int w,
                                 int h) {
   for (int y = 0; y < h; ++y) {
-    memmove(dst, src, w * sizeof(src[0]));
```
yunqingwang1, line 164: Why change this? memmove is good.

avm_dsp/avm_convolve_common.h
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) 2021, Alliance for Open Media. All rights reserved
```
yunqingwang1: Copyright (c) 2026

yunqingwang1:
Looks good.

yunqingwang1 (approved):

avm_dsp/avm_convolve_common.h
jjustiss-apple: Good catch, updating

avm_dsp/avm_convolve.c
jjustiss-apple: This is minor, but this was primarily a semantics change to capture that `memcpy` requires non-aliased buffers, which all the SIMD variants assume already. `dst` is always a prediction buffer and `src` is always a reference frame buffer, so there is no risk of overlap. `memmove`'s contract is weaker on buffer restrictions. 

I fine switching back to `memmove` if you prefer it.

yunqingwang1: ACK

avm_dsp/avm_convolve.c
```diff
@@ -161,7 +153,7 @@ void avm_highbd_convolve_copy_c(const uint16_t *src, ptrdiff_t src_stride,
...
+    memcpy(dst, src, w * sizeof(src[0]));
```
wantehchang, line 156: cc: @jzern 

@jjustiss-apple I'd like to confirm with you that `avm_highbd_convolve_copy_c()` does not need to use `memmove()`.

This function comes from libaom. I am wondering if I should backport your change to libaom.

Thanks!

jjustiss-apple: That was my analysis that memcpy was OK given that `src` is a reference and `dst` is a pred buffer, so we shouldn't have any buffer overlap. So `memcpy`'s stronger restrictions on buffer aliasing apply and should help catch *SAN errors.

PR #5208, opened 2026-08-04
Simplify tcq avx2 functions

jingninghan:
Tcq is only enabled in luma components. Remove unnecessary conditions on plane index.


yeqing-wu (approved):
LGTM

urvangjoshi (approved):

PR #5212, opened 2026-08-05
Bugfix for deblocking filter in partition search: fix distortion scaling

urvangjoshi:
Due to this bug, for sequences with coded bitdepth = 8, the scale of `this_rdc->dist`  was 16x compared to that of `rec_dist`.   Hence, the condition `llabs(this_rdc->dist - rec_dist) < this_rdc->dist / 50` could never be true and so deblocking filter was never activated for this case.

Note:
- Coded bitdepth = 10 was already working correctly -- so no change in output for these cases.
- Coded bitdepth = 12 had this issue as well theoretically, but this is unused in CTC currently.

Results for RA CTC 33 frames at speed 1:
(only affects testsets A3 and A5 that are coded with bitdepth = 8)

```
+------------+--------+--------+---------0--------+--------------+
| Testset | YUV | VMAF | VMAF-NEG | Enc-time |
+------------+--------+--------+---------0--------+--------------+
| A3 | -0.11% | -0.32% | -0.12%  | 99.45% |
| A5 | -0.11% | -0.46% | -0.42%  | 99.76% |
|avg| -0.02% | -0.08% | -0.05%  | 99.41% |
```

STATS_CHANGED for speed >= 1

### DIFF
--- a/av2/encoder/partition_search.c
+++ b/av2/encoder/partition_search.c
@@ -3555,6 +3555,16 @@ static int64_t get_dist_offset_by_deblock(
         mi_row * MI_SIZE >> sub_y, bh >> sub_y);
   }
 
+  const int bit_depth = cm->seq_params.bit_depth;
+  int dist_normal;  // dist is normalized to 16 * 8_bit_content_distortion
+  if (bit_depth <= 10) {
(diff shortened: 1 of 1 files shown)


jianle973 (approved):
Good catch!!!

yeqing-wu (approved):
LGTM

PR #5221, opened 2026-08-06
Refactor write_tile_info_max_tile

wantehchang:
This PR ports the read_tile_params() changes in PR #5187 to the encoder side.

write_tile_info_max_tile() does not need to recalculate width_sb and height_sb. They are equal to tiles->sb_cols and tiles->sb_rows, respectively. See av2_get_tile_limits().

Rename write_tile_info_max_tile() to write_tile_params().

Replace write_tile_syntax_info() by write_tile_params().

### DIFF
--- a/av2/encoder/bitstream.c
+++ b/av2/encoder/bitstream.c
@@ -4216,14 +4216,8 @@ static AVM_INLINE void write_frame_interp_filter(
     avm_wb_write_literal(wb, filter, LOG_SWITCHABLE_FILTERS);
 }
 
-static AVM_INLINE void write_tile_info_max_tile(
-    const CommonTileParams *const tiles, struct avm_write_bit_buffer *wb) {
-  int width_mi = ALIGN_POWER_OF_TWO(tiles->mi_cols, tiles->mib_size_log2);
-  int height_mi = ALIGN_POWER_OF_TWO(tiles->mi_rows, tiles->mib_size_log2);
-  int width_sb = width_mi >> tiles->mib_size_log2;
-  int height_sb = height_mi >> tiles->mib_size_log2;
-  int size_sb, i;
-
+static AVM_INLINE void write_tile_params(const CommonTileParams *const tiles,
+                                         struct avm_write_bit_buffer *wb) {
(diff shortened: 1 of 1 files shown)


av2/encoder/bitstream.c
```diff
@@ -4244,20 +4238,27 @@ static AVM_INLINE void write_tile_info_max_tile(
...
+
     // Explicit tiles with configurable tile widths and heights
     // columns
+    int width_sb = tiles->sb_cols;
     for (i = 0; i < tiles->cols; i++) {
       size_sb = tiles->col_start_sb[i + 1] - tiles->col_start_sb[i];
-      wb_write_uniform(wb, AVMMIN(width_sb, tiles->max_width_sb), size_sb - 1);
+      const int n = AVMMIN(width_sb, tiles->max_width_sb);
+      assert(size_sb >= 1 && size_sb <= n);
+      wb_write_uniform(wb, n, size_sb - 1);
```
wantehchang, line 4250: The new assertion checks that the value we are writing, `size_sb - 1`, is in the range of `0..n-1`.

y-guyon (approved):

av2/encoder/bitstream.c
wantehchang: I am using the assertion to detect a corrupt bitstream. Ideally we should detect this error and call `avm_internal_error()`. But that requires more plumbing. I may try to do that if I have spare time.

More info on this syntax element: This is the syntax element `width_in_sbs_minus_1` in the `tile_params()` syntax structure in Section 5.18.7.3. Tile params syntax:

```
    n = Min(sbCols - startSb, maxTileWidthSb)
    width_in_sbs_minus_1                                   ns(n)
```

The `ns(n)` syntax descriptor is documented as follows:

> Unsigned encoded integer with maximum number of values `n` (i.e., output in range `0..n-1`).

Since we are writing `size_sb - 1` here, it must satisfy the following condition:

```
    size_sb - 1 >= 0 && size_sb - 1 <= n - 1
```

which simplfies to:

```
    size_sb >= 1 && size_sb <= n
```

PR #5224, opened 2026-08-06
av2_set_seq_tile_info,av2_set_tile_info detect err

wantehchang:
Port Yannis's PR #5187 to the encoder, part 2.

Detect errors in av2_set_seq_tile_info() and av2_set_tile_info() with assertions. These two functions return void, and although they have or can be given access to cm->error, some of their callers have not called setjmp(), so they can't call avm_internal_error().

Do not recalculate sb_cols and sb_rows in av2_set_tile_info(). They are already calculated by the av2_get_tile_limits() call and saved in tiles->sb_cols and tiles->sb_rows.

Although av2_set_seq_tile_info() and av2_set_tile_info() are quite similar, they are different enough that it is hard to merge them into a single function.

### DIFF
--- a/av2/encoder/encoder_utils.h
+++ b/av2/encoder/encoder_utils.h
@@ -1007,6 +1007,7 @@ static AVM_INLINE void av2_set_seq_tile_info(SequenceHeader *const seq_params,
       if (j >= tile_cfg->tile_width_count) j = 0;
       start_sb += AVMMIN(size_sb, tiles->max_width_sb);
     }
+    assert(start_sb == sb_cols);
     tiles->cols = i;
     tiles->col_start_sb[i] = sb_cols;
   }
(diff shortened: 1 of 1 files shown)


av2/encoder/encoder_utils.h
```diff
@@ -1037,7 +1039,7 @@ static AVM_INLINE void av2_set_tile_info(AV2_COMMON *const cm,
   CommonTileParams *const tiles = &cm->tiles;
   int i, start_sb;
 
-  av2_get_tile_limits(&cm->tiles, mi_params->mi_rows, mi_params->mi_cols,
+  av2_get_tile_limits(tiles, mi_params->mi_rows, mi_params->mi_cols,
```
wantehchang, line 1042: Here I replaced `&cm->tiles` with the local variable `tiles` (see line 1039).

y-guyon (approved):

urvangjoshi:
@wantehchang :  pls rebase locally and force-push.  Something is stuck on the Github UI.

wantehchang:
Urvang: I just rebased this PR locally and force-pushed it.

PR #5229, opened 2026-08-07
Code cleanup: remove partition arg from pick_sb_modes

urvangjoshi:
This function can only be called with PARTITION_NONE type.


urvangjoshi:
The failures are coming from the head commit (and NOT the PR).  Filed issue: https://github.com/AOMediaCodec/avm/issues/5231

yunqingwang1 (approved):

PR #5230, opened 2026-08-07
set_params_rd_pick_inter_mode: Remove dead code.

urvangjoshi:
This function is only called through the call chain: pick_sb_modes() -->
av2_rd_pick_inter_mode_sb() --> set_params_rd_pick_inter_mode().

In this call chain, `mbmi->partition` is always set to `PARTITION_NONE`.

So, the 2 set of code snippets were dead code.


yunqingwang1 (approved):

yeqing-wu (approved):
LGTM

PR #5236, opened 2026-08-10
Implement non-RD partition evaluation for real-time mode

jianj-g:
This change is bitexact for RTC mode with speed up:

Without Non-RD partition:

```
         Bitrate(kbps)  |  PSNR(Y)  |  PSNR(U)  |  PSNR(V)  |  PSNR(Avg)  |  PSNR(Overall)  |  Encoding time (FPS)
------------------------------------------------------------------------------------------------------------
Summary:    285.057200  |  46.194842  |  44.555717  |  44.840002  |  45.625135    |  45.587741        |    20.9s (14.3 fps)
```

With Non-RD partition:

```
         Bitrate(kbps)  |  PSNR(Y)  |  PSNR(U)  |  PSNR(V)  |  PSNR(Avg)  |  PSNR(Overall)  |  Encoding time (FPS)
------------------------------------------------------------------------------------------------------------
Summary:    285.057200  |  46.194842  |  44.555717  |  44.840002  |  45.625135    |  45.587741        |    18.4s (16.2 fps)
```

Tested with command line:

```
./avmenc --passes=1 --bit-depth=8 --obu --lag-in-frames=0 --kf-min-dist=1000 --kf-max-dist=1000 screen_inp.1280_720.yuv -w 1280 -h 720 -o out.ivf --fps=15/1 --target-bitrate=400 --auto-alt-ref=0 --enable-tpl-model=0 --enable-keyframe-filtering=0 --enable-deblocking=1 --enable-restoration=0  --enable-intra-edge-filter=0 --enable-flip-idtx=1 --enable-masked-comp=0 --enable-onesided-comp=0 --enable-interintra-comp=0 --enable-smooth-interintra=0 --enable-diff-wtd-comp=0 --enable-interinter-wedge=0 --enable-interintra-wedge=0 --enable-global-motion=0 --enable-warped-motion=0 --enable-smooth-intra=0 --enable-paeth-intra=0 --enable-cfl-intra=0 --force-video-mode=1 --enable-overlay=0 --enable-angle-delta=0 --enable-trellis-quant=0 --enable-qm=0 --use-intra-dct-only=1 --coeff-cost-upd-freq=2 --mode-cost-upd-freq=2 --mv-cost-upd-freq=3 --frame-parallel=0 --aq-mode=0 --deltaq-mode=0 --frame-boost=0 --noise-sensitivity=0 --cdf-update-mode=1 --reduced-reference-set=1 --enable-ref-frame-mvs=0 --erp-pruning-level=0 --use-ml-erp-pruning=0 --enable-ext-partitions=0 --enable-mrls=0 --enable-pc-wiener=0 --enable-wiener-nonsep=0 --enable-tip=0 --enable-bawp=0 --enable-cwp=0 --enable-imp-msk-bld=0 --enable-ist=0 --enable-inter-ist=0 --enable-inter-ddt=0 --enable-cctx=0 --enable-ibp=0 --max-drl-refmvs=0 --max-drl-refbvs=0 --enable-refmvbank=0 --enable-opfl-refine=0 --enable-ccso=0 --enable-lf-sub-pu=0 --enable-adaptive-mvd=0 --enable-flex-mvres=0 --enable-joint-mvd=0 --enable-refinemv=0 --enable-mvd-sign-derive=0 --enable-parity-hiding=0 --enable-warp-delta=0 --enable-warp-extend=0 --tile-columns=0 --gf-min-pyr-height=0 --gf-max-pyr-height=0 --end-usage=q --use-fixed-qp-offsets=0 --psnr --threads=1 --enable-sdp=0 --min-gf-interval=8 --max-gf-interval=8 --explicit-ref-frame-map=1 --enable-bru=0 --enable-ext-seg=0 --enable-six-param-warp-delta=0 --enable-cdef=1 --enable-cdef-on-skip-txfm=0 --limit=1000 --dpb-size=3 --use-inter-dct-only=0 enable-fsc=1 --enable-idtx-intra=1 --enable-palette=1 --enable-intrabc-ext=0 --enable-intrabc=1 --tune-content=1 --rt --cpu-used=6 --min-qp=100 --max-qp=100 --enable-gdf=0 --max-reference-frames=1
```

### DIFF
--- a/av2/encoder/partition_search.c
+++ b/av2/encoder/partition_search.c
@@ -2356,6 +2356,193 @@ static void init_partition_costs(const AV2_COMMON *const cm,
   }
 }
 
+/*!\brief Fast non-RD block partition encoding.
+ *
+ * \ingroup partition_search
+ * Encodes a block using pre-calculated partition patterns (e.g. from
+ * variance-based partitioning) without executing full Rate-Distortion (RD)
+ * searches across all candidate partition types. Recursively traverses basic
+ * partition types: PARTITION_NONE, PARTITION_HORZ, PARTITION_VERT, and
+ * PARTITION_SPLIT.
+ *
+ * \param[in]    cpi         Top-level encoder structure
+ * \param[in]    td          Pointer to thread data
+ * \param[in]    tile_data   Pointer to tile-level encoder data structure
+ * \param[in]    mib         Array of MB_MODE_INFO pointers starting from the
+ * current block
+ * \param[in]    tp          Pointer to the starting token
+ * \param[in]    mi_row      Row coordinate of the block (in MI units)
+ * \param[in]    mi_col      Column coordinate of the block (in MI units)
+ * \param[in]    bsize       Current block size
+ * \param[in]    ptree       Pointer to PARTITION_TREE node holding
+ * pre-calculated partitions
+ * \param[in]    pc_tree     Pointer to PC_TREE node holding picked partition
+ * information
+ * \param[in]    ptree_luma  Pointer to luma partition tree (used for chroma
+ * partition estimation)
+ */
+void av2_nonrd_use_partition(AV2_COMP *cpi, ThreadData *td,
+                             TileDataEnc *tile_data, MB_MODE_INFO **mib,
+                             TokenExtra **tp, int mi_row, int mi_col,
+                             BLOCK_SIZE bsize, PARTITION_TREE *ptree,
+                             PC_TREE *pc_tree, PARTITION_TREE *ptree_luma) {
+  AV2_COMMON *const cm = &cpi->common;
+  const CommonModeInfoParams *const mi_params = &cm->mi_params;
+  if (mi_row >= mi_params->mi_rows || mi_col >= mi_params->mi_cols) return;
+
+  TileInfo *const tile_info = &tile_data->tile_info;
+  MACROBLOCK *const x = &td->mb;
+  MACROBLOCKD *const xd = &x->e_mbd;
+  const int ss_x = xd->plane[1].subsampling_x;
+  const int ss_y = xd->plane[1].subsampling_y;
+  assert(bsize < BLOCK_SIZES_ALL);
+  const int hbh = mi_size_high[bsize] / 2;
+  const int hbw = mi_size_wide[bsize] / 2;
+  const PARTITION_TYPE partition =
+      get_preset_partition(cm, xd->tree_type, mi_row, mi_col, bsize, ptree);
+  const BLOCK_SIZE subsize = get_partition_subsize(bsize, partition);
+  RD_STATS this_rdc, best_rdc;
+  int rate;
+  av2_invalid_rd_stats(&this_rdc);
+  av2_invalid_rd_stats(&best_rdc);
+
+  if (!frame_is_intra_only(cm))
+    pc_tree->region_type = MIXED_INTER_INTRA_REGION;
+  else
+    pc_tree->region_type = INTRA_REGION;
+  REGION_TYPE cur_region_type = pc_tree->region_type;
+  if (is_inter_sdp_chroma(cm, cur_region_type, x->e_mbd.tree_type)) {
+    if (pc_tree->none_chroma == NULL) {
+      pc_tree->none_chroma =
+          av2_alloc_pmc(cm, xd->tree_type, mi_row, mi_col, bsize, pc_tree,
+                        PARTITION_NONE, 0, ss_x, ss_y, &td->shared_coeff_buf);
+    }
+  } else {
+    if (pc_tree->none[cur_region_type] == NULL) {
+      pc_tree->none[cur_region_type] =
+          av2_alloc_pmc(cm, xd->tree_type, mi_row, mi_col, bsize, pc_tree,
+                        PARTITION_NONE, 0, ss_x, ss_y, &td->shared_coeff_buf);
+    }
+  }
+
+  PICK_MODE_CONTEXT *ctx_none =
+      is_inter_sdp_chroma(cm, cur_region_type, x->e_mbd.tree_type)
+          ? pc_tree->none_chroma
+          : pc_tree->none[pc_tree->region_type];
+
+  pc_tree->partitioning = partition;
+
+  if (bsize == BLOCK_16X16 && cpi->vaq_refresh) {
+    av2_set_offsets(cpi, tile_info, x, mi_row, mi_col, bsize,
+                    &pc_tree->chroma_ref_info);
+    x->mb_energy = av2_log_block_var(cpi, x, bsize
+#if CONFIG_MIXED_LOSSLESS_ENCODE
+                                     ,
+                                     mi_row, mi_col
+#endif  // CONFIG_MIXED_LOSSLESS_ENCODE
+    );
+  }
+
+  if (bsize == cm->sb_size) {
+    if (pc_tree)
+      pc_tree->is_cfl_allowed_for_this_chroma = CFL_DISALLOWED_FOR_CHROMA;
+
+    xd->is_cfl_allowed_in_sdp =
+        is_cfl_allowed_for_sdp(cm, xd, ptree_luma, PARTITION_NONE, bsize);
+    ptree->is_cfl_allowed_for_this_chroma_partition = CFL_DISALLOWED_FOR_CHROMA;
+  }
+
+  if (partition == PARTITION_NONE) {
+    xd->is_cfl_allowed_in_sdp =
+        pc_tree->is_cfl_allowed_for_this_chroma |
+        ptree->is_cfl_allowed_for_this_chroma_partition |
+        is_cfl_allowed_for_sdp(cm, xd, ptree_luma, partition, bsize);
+
+  } else {
+    pc_tree->is_cfl_allowed_for_this_chroma =
+        ((pc_tree->parent) ? pc_tree->parent->is_cfl_allowed_for_this_chroma
+                           : 0) |
+        is_cfl_allowed_for_sdp(cm, xd, ptree_luma, partition, bsize);
+    ptree->is_cfl_allowed_for_this_chroma_partition = CFL_DISALLOWED_FOR_CHROMA;
+  }
+
+  switch (partition) {
+    case PARTITION_NONE:
+      pick_sb_modes(cpi, td, tile_data, x, mi_row, mi_col, &this_rdc,
+                    pc_tree->region_type, pc_tree->sb_root_partition_info,
+                    bsize, ctx_none, best_rdc);
+      encode_b(cpi, tile_data, td, tp, mi_row, mi_col, OUTPUT_ENABLED, subsize,
+               partition, ctx_none, &rate);
+      break;
+    case PARTITION_HORZ:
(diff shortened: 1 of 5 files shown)


leolzhao:
Please provide the results for this pull request. @jianj-g

jianj-g:
> Please provide the results for this pull request. @jianj-g

Updated the PR description with the result. This PR is bitexact for RTC mode with speed up.

av2/encoder/partition_search.c
```diff
@@ -2356,6 +2356,199 @@ static void init_partition_costs(const AV2_COMMON *const cm,
...
+ * \param[in]    mi_col      Column coordinate of the block (in MI units)
+ * \param[in]    bsize       Current block size
+ * \param[in]    ptree       Pointer to PARTITION_TREE node holding
+ * pre-calculated partitions
+ * \param[in]    pc_tree     Pointer to PC_TREE node holding picked partition
+ * information
+ * \param[in]    ptree_luma  Pointer to luma partition tree (used for chroma
+ * partition estimation)
+ */
+void av2_nonrd_use_partition(AV2_COMP *cpi, ThreadData *td,
```
urvangjoshi, line 2384: This new function seem almost same as `av2_rd_use_partition`.  What's the difference exactly?   
Also,
- Why is there no stat change, but speed-up in RTC mode with this change?
- Why can't this function be used in good speed mode?

leolzhao: I have the same question with @urvangjoshi. There is also no RDO search in function `av2_rd_use_partition`, so what's the reason for the run-time saving? @jianj-g

jianj-g: This new function doesn't execute the following compared with `av2_rd_use_partition`:

1. It doesn't use `encode_sb`. After mode selection, it directly calls `encode_b` to tokenize.
2. There's no rate/distortion calculation / accumulation.
3. There's no syntax cost calculation (`init_partition_cost`).

Good mode requires RD optimization to pick the best partition. In RTC mode as it is, it uses the pre-determined partition (by varianced based partition in the last PR).

Since the partitioning is already determined in RTC mode, for leaf blocks, `pick_sb_modes()` and token encoding produces identical motion vectors, transform coefficients, and bitstream syntax. So RTC mode is bitexact. But this new function avoids all the extra calculations mentioned above, so it's faster.

leolzhao: Then, is the rate/distortion calculation/accumulation redundant in av2_rd_use_partition function? If so, can we directly modify the code in function av2_rd_use_partition instead of adding a new function?

jianj-g: No, it's not redundant. The rate/distortion in `av2_rd_use_partition` is used later in the encoding in non-RTC mode... so removing them would break non-RTC path...

We **could** modify `av2_rd_use_partition` but IMHO it'll be less cleaner and make the code more complicated and less readable. They may look similar because traversing the partitioning tree looks similar in  both functions but `av2_rd_use_partition` does more than that.

Having a dedicated non-rd function here helps keeping a clear separation between different modes and it's easier to strip off unnecessary code when needed, for example, we added `CONFIG_REALTIME_ONLY` in libaom to reduce the binary size for RTC focused applications.

leolzhao: Thanks for the explanation, and it makes sense to me to have a different function.

leolzhao (approved):

av2/encoder/partition_search.c
jianj-g: Thanks for the review!

av2/encoder/partition_search.c
```diff
@@ -2356,6 +2356,199 @@ static void init_partition_costs(const AV2_COMMON *const cm,
...
+  PICK_MODE_CONTEXT *ctx_none =
+      is_inter_sdp_chroma(cm, cur_region_type, x->e_mbd.tree_type)
+          ? pc_tree->none_chroma
+          : pc_tree->none[pc_tree->region_type];
+
+  if (mi_row >= mi_params->mi_rows || mi_col >= mi_params->mi_cols) return;
+
+  pc_tree->partitioning = partition;
+
+  av2_save_context(x, &x_ctx, mi_row, mi_col, bsize, num_planes);
```
yeqing-wu: Within this function, we did not restore the ctx from x_ctx, then why do we need to store the ctx to x_ctx? save_context is a non-trivial memcpy of neighbor arrays.

av2/encoder/partition_search.c
```diff
@@ -2356,6 +2356,199 @@ static void init_partition_costs(const AV2_COMMON *const cm,
...
+                        PARTITION_NONE, 0, ss_x, ss_y, &td->shared_coeff_buf);
+    }
+  }
+
+  PICK_MODE_CONTEXT *ctx_none =
+      is_inter_sdp_chroma(cm, cur_region_type, x->e_mbd.tree_type)
+          ? pc_tree->none_chroma
+          : pc_tree->none[pc_tree->region_type];
+
+  if (mi_row >= mi_params->mi_rows || mi_col >= mi_params->mi_cols) return;
```
yeqing-wu: should we move this check at the beginning of this function so that we can avoid lots of check, initialization, memory allocation in the case it should return?

urvangjoshi (approved):

av2/encoder/partition_search.c
```diff
@@ -2356,6 +2356,199 @@ static void init_partition_costs(const AV2_COMMON *const cm,
...
+  TileInfo *const tile_info = &tile_data->tile_info;
+  MACROBLOCK *const x = &td->mb;
+  MACROBLOCKD *const xd = &x->e_mbd;
+  const int ss_x = xd->plane[1].subsampling_x;
+  const int ss_y = xd->plane[1].subsampling_y;
+  assert(bsize < BLOCK_SIZES_ALL);
+  const int bs = mi_size_wide[bsize];
+  const int hbs = bs / 2;
+  const int hbh = mi_size_high[bsize] / 2;
+  const int hbw = mi_size_wide[bsize] / 2;
```
yeqing-wu, line 2400: hbw = hbs?

yeqing-wu: I think we do not need bs, hbs by directly using the following code, and replace all code with bs, hbs with hbw, right?
```
const int hbh = mi_size_high[bsize] / 2;
const int hbw = mi_size_wide[bsize] / 2;
```

av2/encoder/partition_search.c
```diff
@@ -2356,6 +2356,199 @@ static void init_partition_costs(const AV2_COMMON *const cm,
...
+          PARTITION_HORZ, 0, 0, ss_x, ss_y);
+      pc_tree->horizontal[cur_region_type][1] = av2_alloc_pc_tree_node(
+          xd->tree_type, mi_row + hbh, mi_col, cm->sb_size, subsize, pc_tree,
+          PARTITION_HORZ, 1, 1, ss_x, ss_y);
+
+      av2_nonrd_use_partition(cpi, td, tile_data, mib, tp, mi_row, mi_col,
+                              subsize, ptree ? ptree->sub_tree[0] : NULL,
+                              pc_tree->horizontal[cur_region_type][0],
+                              get_partition_subtree_const(ptree_luma, 0));
+      if (bsize >= BLOCK_8X8 && mi_row + hbs < mi_params->mi_rows) {
```
yeqing-wu: hbs is for width, but here the code is mi_row, thus it should be a bug here, right? Should not it be `mi_row + hbh < mi_params->mi_rows`?

```
  const int bs = mi_size_wide[bsize];
  const int hbs = bs / 2;
```

yeqing-wu (commented):
I am still reviewing this, here are partial comments.

av2/encoder/partition_search.c
```diff
@@ -2356,6 +2356,199 @@ static void init_partition_costs(const AV2_COMMON *const cm,
...
+                                mi_col + hbw, subsize,
+                                ptree ? ptree->sub_tree[1] : NULL,
+                                pc_tree->vertical[cur_region_type][1],
+                                get_partition_subtree_const(ptree_luma, 1));
+      }
+      break;
+    case PARTITION_SPLIT:
+      for (int i = 0; i < SUB_PARTITIONS_SPLIT; i++) {
+        int x_idx = (i & 1) * hbs;
+        int y_idx = (i >> 1) * hbs;
```
yeqing-wu: both dimensions use hbs, is it a bug?

av2/encoder/partition_search.c
```diff
@@ -2356,6 +2356,199 @@ static void init_partition_costs(const AV2_COMMON *const cm,
...
+  const int hbh = mi_size_high[bsize] / 2;
+  const int hbw = mi_size_wide[bsize] / 2;
+  const PARTITION_TYPE partition =
+      get_preset_partition(cm, xd->tree_type, mi_row, mi_col, bsize, ptree);
+  const BLOCK_SIZE subsize = get_partition_subsize(bsize, partition);
+  RD_SEARCH_MACROBLOCK_CONTEXT x_ctx;
+  RD_STATS this_rdc, best_rdc;
+  int rate;
+  av2_invalid_rd_stats(&this_rdc);
+  av2_invalid_rd_stats(&best_rdc);
```
yeqing-wu, line 2407: we never use best_rdc and this_rdc, are they dead code?

av2/encoder/partition_search.c
```diff
@@ -2356,6 +2356,199 @@ static void init_partition_costs(const AV2_COMMON *const cm,
...
+  const int bs = mi_size_wide[bsize];
+  const int hbs = bs / 2;
+  const int hbh = mi_size_high[bsize] / 2;
+  const int hbw = mi_size_wide[bsize] / 2;
+  const PARTITION_TYPE partition =
+      get_preset_partition(cm, xd->tree_type, mi_row, mi_col, bsize, ptree);
+  const BLOCK_SIZE subsize = get_partition_subsize(bsize, partition);
+  RD_SEARCH_MACROBLOCK_CONTEXT x_ctx;
+  RD_STATS this_rdc, best_rdc;
+  int rate;
```
yeqing-wu, line 2405: rate is only read, is it also the dead code? If not, should we initialize it?

yeqing-wu (commented):
Thanks for supporting this feature, here are all of my review comments.

yunqingwang1 (approved):

av2/encoder/partition_search.c
jianj-g: `best_rdc` is input parameter for `pick_sb_modes` (we fill it up with `av2_invalid_rd_stats` meaning there's no upper bound rd threshold.

`this_rdc` is output parameter for `pick_sb_modes` but this is redundant - it's never used but it's required as an argument to `pick_sb_modes`

av2/encoder/partition_search.c
jianj-g: similarly `rate` is also a dummy output parameter for `encode_b`

av2/encoder/partition_search.c
jianj-g: Good catch!

av2/encoder/partition_search.c
jianj-g: Yes it is! Thanks for catching this!

yeqing-wu (approved):
Thanks. LGTM

urvangjoshi:
(Merging without waiting for CI, on behalf of Yunqing)

PR #5246, opened 2026-08-11
Force max pb aspect ratio to be 4 at speed 4

motongxu:
max_pb_aspect_ratio is set to 4 at speed 4.

Anchor: https://github.com/AOMediaCodec/avm/commit/e2810080c0208f323183e58126c3fee1802a7ac6
Tests: cpu-used =4, RA
```
+--------------------------+---------+---------+---------+---------+----------+----------+
| Summary                  | Y       | U       | V       | YUV     | Enc-time | Dec-time |
+--------------------------+---------+---------+---------+---------+----------+----------+
| Class A1_4K              | 0.09%   | -0.06%  | -0.99%  | 0.03%   | 96.4%    | 99.5%    |
| Class A2_2K              | 0.09%   | -0.35%  | -0.20%  | 0.06%   | 97.7%    | 99.6%    |
| Class A3_720p            | 0.10%   | -0.43%  | 0.02%   | 0.08%   | 93.8%    | 100%     |
| Class A4_360p            | 0.23%   | -1.46%  | 1.21%   | 0.17%   | 95.0%    | 101%     |
| Class A5_270p            | 0.00%   | 0.00%   | 0.00%   | 0.00%   | 100.0%   | 100%     |
| Class B1_SYN             | 0.21%   | 0.48%   | 0.21%   | 0.21%   | 95.4%    | 101%     |
| Overall Mandatory w/o B2 | 0.14%   | #VALUE! | 0.29%   | 0.13%   | 95.5%    | 99%      |
| Class B2_SCC             | 0.00%   | 0.00%   | 0.00%   | 0.00%   | 100.0%   | 100%     |
+--------------------------+---------+---------+---------+---------+----------+----------+
```

### DIFF
--- a/av2/encoder/partition_search.c
+++ b/av2/encoder/partition_search.c
@@ -3933,16 +3933,10 @@ static void none_partition_search(
   // Check if partition none is allowed.
   const int bw = block_size_wide[bsize];
   const int bh = block_size_high[bsize];
-  if (cpi->sf.part_sf.force_max_pb_aspect_ratio) {
-    cpi->common.seq_params.max_pb_aspect_ratio_log2_m1 =
-        cpi->sf.part_sf.force_max_pb_aspect_ratio == 2
-            ? 0
-            : (cpi->sf.part_sf.force_max_pb_aspect_ratio == 4 ? 1 : 2);
-  }
   const int max_partition_aspect_ratio =
-      1 << (cm->seq_params.max_pb_aspect_ratio_log2_m1 + 1);
-  assert(max_partition_aspect_ratio == 2 || max_partition_aspect_ratio == 4 ||
-         max_partition_aspect_ratio == 8);
+      cpi->sf.part_sf.force_max_pb_aspect_ratio
+          ? cpi->sf.part_sf.force_max_pb_aspect_ratio
+          : 1 << (cm->seq_params.max_pb_aspect_ratio_log2_m1 + 1);
   const bool invalid_aspect_ratio = bw > bh * max_partition_aspect_ratio ||
                                     bh > bw * max_partition_aspect_ratio;
   if (invalid_aspect_ratio) return;
--- a/av2/encoder/partition_search.c
+++ b/av2/encoder/partition_search.c
@@ -3934,9 +3934,7 @@ static void none_partition_search(
   const int bw = block_size_wide[bsize];
   const int bh = block_size_high[bsize];
   const int max_partition_aspect_ratio =
-      cpi->sf.part_sf.force_max_pb_aspect_ratio
-          ? cpi->sf.part_sf.force_max_pb_aspect_ratio
-          : 1 << (cm->seq_params.max_pb_aspect_ratio_log2_m1 + 1);
+      1 << (cm->seq_params.max_pb_aspect_ratio_log2_m1 + 1);
   const bool invalid_aspect_ratio = bw > bh * max_partition_aspect_ratio ||
                                     bh > bw * max_partition_aspect_ratio;
   if (invalid_aspect_ratio) return;
--- a/av2/encoder/partition_search.c
+++ b/av2/encoder/partition_search.c
@@ -3933,8 +3933,16 @@ static void none_partition_search(
   // Check if partition none is allowed.
   const int bw = block_size_wide[bsize];
   const int bh = block_size_high[bsize];
+  if (cpi->sf.part_sf.force_max_pb_aspect_ratio) {
+    cpi->common.seq_params.max_pb_aspect_ratio_log2_m1 =
+        cpi->sf.part_sf.force_max_pb_aspect_ratio == 2
+            ? 0
+            : (cpi->sf.part_sf.force_max_pb_aspect_ratio == 4 ? 1 : 2);
+  }
   const int max_partition_aspect_ratio =
       1 << (cm->seq_params.max_pb_aspect_ratio_log2_m1 + 1);
+  assert(max_partition_aspect_ratio == 2 || max_partition_aspect_ratio == 4 ||
+         max_partition_aspect_ratio == 8);
   const bool invalid_aspect_ratio = bw > bh * max_partition_aspect_ratio ||
                                     bh > bw * max_partition_aspect_ratio;
   if (invalid_aspect_ratio) return;
--- a/av2/encoder/speed_features.c
+++ b/av2/encoder/speed_features.c
@@ -1110,10 +1110,11 @@ void av2_set_speed_features_framesize_dependent(AV2_COMP *cpi, int speed) {
       !cpi->seq_params_locked) {
     sf->part_sf.force_max_pb_aspect_ratio = 4;
     if (sf->part_sf.force_max_pb_aspect_ratio) {
+      const int new_max_ratio =
+          AVMMIN(sf->part_sf.force_max_pb_aspect_ratio,
+                 oxcf->part_cfg.max_partition_aspect_ratio);
       cpi->common.seq_params.max_pb_aspect_ratio_log2_m1 =
-          sf->part_sf.force_max_pb_aspect_ratio == 2
-              ? 0
-              : (sf->part_sf.force_max_pb_aspect_ratio == 4 ? 1 : 2);
+          new_max_ratio == 2 ? 0 : (new_max_ratio == 4 ? 1 : 2);
     }
   }
 }
--- a/av2/encoder/speed_features.c
+++ b/av2/encoder/speed_features.c
@@ -579,6 +579,7 @@ static void set_good_speed_features_framesize_independent(
 
     sf->part_sf.simple_motion_search_prune_agg = 2;
     sf->part_sf.simple_motion_search_reduce_search_steps = 4;
+    sf->part_sf.force_max_pb_aspect_ratio = 4;
 
     sf->inter_sf.alt_ref_search_fp = 1;
     sf->inter_sf.txfm_rd_gate_level = boosted ? 0 : 4;
@@ -791,7 +792,9 @@ static AVM_INLINE void init_part_sf(PARTITION_SPEED_FEATURES *part_sf) {
 #endif  // CONFIG_ML_PART_SPLIT
   part_sf->disable_ext_partitions = false;
   part_sf->disable_uneven_4way_partitions = false;
+
   part_sf->disable_extended_sdp = false;
+  part_sf->force_max_pb_aspect_ratio = 0;
 }
 
 static AVM_INLINE void init_mv_sf(MV_SPEED_FEATURES *mv_sf) {
--- a/av2/encoder/speed_features.c
+++ b/av2/encoder/speed_features.c
@@ -579,9 +579,6 @@ static void set_good_speed_features_framesize_independent(
 
     sf->part_sf.simple_motion_search_prune_agg = 2;
     sf->part_sf.simple_motion_search_reduce_search_steps = 4;
-    if (!cpi->is_screen_content_type) {
-      sf->part_sf.force_max_pb_aspect_ratio = 4;
-    }
 
     sf->inter_sf.alt_ref_search_fp = 1;
     sf->inter_sf.txfm_rd_gate_level = boosted ? 0 : 4;
@@ -1086,6 +1083,8 @@ static AVM_INLINE void set_erp_speed_features_framesize_dependent(
 void av2_set_speed_features_framesize_dependent(AV2_COMP *cpi, int speed) {
   SPEED_FEATURES *const sf = &cpi->sf;
   const AV2EncoderConfig *const oxcf = &cpi->oxcf;
+  const int is_270p_or_lesser =
+      AVMMIN(cpi->common.width, cpi->common.height) <= 270;
 
   if (oxcf->mode == GOOD) {
     set_good_speed_feature_framesize_dependent(cpi, sf, speed);
@@ -1105,6 +1104,17 @@ void av2_set_speed_features_framesize_dependent(AV2_COMP *cpi, int speed) {
 
   if (oxcf->txfm_cfg.enable_tx_partition == 0) {
     sf->tx_sf.enable_tx_partition = false;
+  }
+
+  if (speed >= 4 && !cpi->is_screen_content_type && !is_270p_or_lesser &&
+      !cpi->seq_params_locked) {
+    sf->part_sf.force_max_pb_aspect_ratio = 4;
+    if (sf->part_sf.force_max_pb_aspect_ratio) {
+      cpi->common.seq_params.max_pb_aspect_ratio_log2_m1 =
+          sf->part_sf.force_max_pb_aspect_ratio == 2
+              ? 0
+              : (sf->part_sf.force_max_pb_aspect_ratio == 4 ? 1 : 2);
+    }
   }
 }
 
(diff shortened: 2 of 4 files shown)


av2/encoder/partition_search.c
```diff
@@ -3757,8 +3757,16 @@ static void none_partition_search(
   // Check if partition none is allowed.
   const int bw = block_size_wide[bsize];
   const int bh = block_size_high[bsize];
+  if (cpi->sf.part_sf.force_max_pb_aspect_ratio) {
+    cpi->common.seq_params.max_pb_aspect_ratio_log2_m1 =
```
urvangjoshi: You should NOT be changing seq_params  during partition search.  ` cpi->common.seq_params.max_pb_aspect_ratio_log2_m1`  should be set in the beginning of the sequence (only once).

av2/encoder/partition_search.c
```diff
@@ -3757,8 +3757,16 @@ static void none_partition_search(
...
+        cpi->sf.part_sf.force_max_pb_aspect_ratio == 2
+            ? 0
+            : (cpi->sf.part_sf.force_max_pb_aspect_ratio == 4 ? 1 : 2);
+  }
   const int max_partition_aspect_ratio =
       1 << (cm->seq_params.max_pb_aspect_ratio_log2_m1 + 1);
+  assert(max_partition_aspect_ratio == 2 || max_partition_aspect_ratio == 4 ||
```
urvangjoshi: You don't need this assert here,  as it's already checked earlier: https://github.com/AOMediaCodec/avm/blob/5c6aa87a3206563a54a667b7b48e270e09ad3227/av2/common/av2_common_int.h#L4869

urvangjoshi:
The loss on A2 is on higher side -- but ratio maybe acceptable.  

But can you share results for other sets?  I'm worried that loss maybe even higher for lower resolutions and also for B1 / B2.

yeqing-wu (approved):
LGTM

leolzhao:
> The loss on A2 is on higher side -- but ratio maybe acceptable.
> 
> But can you share results for other sets? I'm worried that loss maybe even higher for lower resolutions and also for B1 / B2.

We didn't test class B1 and B2. We can turn off this speed feature for screen content by checking !cpi->is_screen_content_type. What do you think? @urvangjoshi

av2/encoder/partition_search.c
```diff
@@ -3758,7 +3758,9 @@ static void none_partition_search(
   const int bw = block_size_wide[bsize];
   const int bh = block_size_high[bsize];
   const int max_partition_aspect_ratio =
-      1 << (cm->seq_params.max_pb_aspect_ratio_log2_m1 + 1);
+      cpi->sf.part_sf.force_max_pb_aspect_ratio
```
urvangjoshi: The updated code still doesn't look correct to me.    You are only forcing the aspect ratio during partition search, but you are NOT modifying the `max_pb_aspect_ratio_log2_m1`  that's signaled at sequence level.

Changing the signaled `max_pb_aspect_ratio_log2_m1` will ensure minimal coding loss.

leolzhao: This speed feature is doing the same strategy as sf->part_sf.disable_uneven_4way_partitions, which is also turning off uneven 4 way by encoder only method rather than changing the signaling.

I recall you also did the test with changing the command line flag, is there a big difference about the coding performance? @motongxu

urvangjoshi: > This speed feature is doing the same strategy as sf->part_sf.disable_uneven_4way_partitions, which is also turning off uneven 4 way by encoder only method rather than changing the signaling.

We should change that too, if that is the case.  I'll look into this one shortly.

urvangjoshi: PR https://github.com/AOMediaCodec/avm/pull/5254 implements this change for `disable_uneven_4way_partitions` and `disable_ext_partitions` speed features.

@motongxu :  Pls use a similar implmentation for this PR as well.

> I recall you also did the test with changing the command line flag, is there a big difference about the coding performance? @motongxu

I'd expect a bigger difference in All-Intra,  even if difference is small in RA.
So, let's go with the normative signaling option for general coding efficiency.

urvangjoshi:
> > The loss on A2 is on higher side -- but ratio maybe acceptable.
> > But can you share results for other sets? I'm worried that loss maybe even higher for lower resolutions and also for B1 / B2.
> 
> We didn't test class B1 and B2. We can turn off this speed feature for screen content by checking !cpi->is_screen_content_type. What do you think? @urvangjoshi

We should decide based on data.   Perhaps, run B2 set and A5 set RA 33 frames at least (should be quick)  and decide based on the results if speed feature should be disabled for low-res and/or screen content.

leolzhao:
> > > The loss on A2 is on higher side -- but ratio maybe acceptable.
> > > But can you share results for other sets? I'm worried that loss maybe even higher for lower resolutions and also for B1 / B2.
> > 
> > 
> > We didn't test class B1 and B2. We can turn off this speed feature for screen content by checking !cpi->is_screen_content_type. What do you think? @urvangjoshi
> 
> We should decide based on data. Perhaps, run B2 set and A5 set RA 33 frames at least (should be quick) and decide based on the results if speed feature should be disabled for low-res and/or screen content.

Agreed. Could you please share the results on class A5 and B2? @motongxu

av2/encoder/partition_search.c
motongxu: Thank you all for the comments and suggestions. I have updated the implementation accordingly in the latest version.

motongxu:
> > > The loss on A2 is on higher side -- but ratio maybe acceptable.
> > > But can you share results for other sets? I'm worried that loss maybe even higher for lower resolutions and also for B1 / B2.
> > 
> > 
> > We didn't test class B1 and B2. We can turn off this speed feature for screen content by checking !cpi->is_screen_content_type. What do you think? @urvangjoshi
> 
> We should decide based on data. Perhaps, run B2 set and A5 set RA 33 frames at least (should be quick) and decide based on the results if speed feature should be disabled for low-res and/or screen content.

Test results of A3-B1 are updated in the description (the anchor's UPSNR of Vertical_Bayshore_270x480_2997 is not monotonic so the U result of class A5 is not available).
B2 sequences are also tested but the trade off did not meet the bar so I have disabled it for screen content. B2 results are as follows for your reference.
```
+--------------------------+---------+---------+---------+---------+----------+----------+
| Summary                  | Y       | U       | V       | YUV     | Enc-time | Dec-time |
+--------------------------+---------+---------+---------+---------+----------+----------+
| Class B2_SCC             | 0.66%   | 1.04%   | 1.03%   | 0.71%   | 97.2%    | 101%     |
+--------------------------+---------+---------+---------+---------+----------+----------+
```

av2/encoder/speed_features.c
```diff
@@ -1221,6 +1225,13 @@ void av2_set_speed_features_framesize_independent(AV2_COMP *cpi, int speed) {
   if (!cpi->seq_params_locked) {
     cpi->common.seq_params.enable_restoration &= !sf->lpf_sf.disable_lr_filter;
 
+    if (sf->part_sf.force_max_pb_aspect_ratio) {
+      cpi->common.seq_params.max_pb_aspect_ratio_log2_m1 =
```
urvangjoshi: `max_pb_aspect_ratio_log2_m1`  may have been set already  from command-line by the user.   You need to add a condition to not change user-provided value.

How about this?   (Pls double check the code)

```
    if (sf->part_sf.force_max_pb_aspect_ratio) {
      const int new_max_pb_aspect_ratio_log2_m1 =
          sf->part_sf.force_max_pb_aspect_ratio == 2
              ? 0
              : (sf->part_sf.force_max_pb_aspect_ratio == 4 ? 1 : 2);
      cpi->common.seq_params.max_pb_aspect_ratio_log2_m1 = AVMMAX(cpi->common.seq_params.max_pb_aspect_ratio_log2_m1, new_max_pb_aspect_ratio_log2_m1);
    }
```

urvangjoshi:
> > > > The loss on A2 is on higher side -- but ratio maybe acceptable.
> > > > But can you share results for other sets? I'm worried that loss maybe even higher for lower resolutions and also for B1 / B2.
> > > 
> > > 
> > > We didn't test class B1 and B2. We can turn off this speed feature for screen content by checking !cpi->is_screen_content_type. What do you think? @urvangjoshi
> > 
> > 
> > We should decide based on data. Perhaps, run B2 set and A5 set RA 33 frames at least (should be quick) and decide based on the results if speed feature should be disabled for low-res and/or screen content.
> 
> Test results of A3-B1 are updated in the description (the anchor's UPSNR of Vertical_Bayshore_270x480_2997 is not monotonic so the U result of class A5 is not available). B2 sequences are also tested but the trade off did not meet the bar so I have disabled it for screen content. B2 results are as follows for your reference.
> 
> ```
> +--------------------------+---------+---------+---------+---------+----------+----------+
> | Summary                  | Y       | U       | V       | YUV     | Enc-time | Dec-time |
> +--------------------------+---------+---------+---------+---------+----------+----------+
> | Class B2_SCC             | 0.66%   | 1.04%   | 1.03%   | 0.71%   | 97.2%    | 101%     |
> +--------------------------+---------+---------+---------+---------+----------+----------+
> ```

Thanks for the update!

Based on the results,  we should also disable the speed feature for <= 270p.
I have done something similar for another speed feature:  https://github.com/AOMediaCodec/avm/blob/av2-enc/av2/encoder/speed_features.c#L1053

motongxu:
> > > > > The loss on A2 is on higher side -- but ratio maybe acceptable.
> > > > > But can you share results for other sets? I'm worried that loss maybe even higher for lower resolutions and also for B1 / B2.
> > > > 
> > > > 
> > > > We didn't test class B1 and B2. We can turn off this speed feature for screen content by checking !cpi->is_screen_content_type. What do you think? @urvangjoshi
> > > 
> > > 
> > > We should decide based on data. Perhaps, run B2 set and A5 set RA 33 frames at least (should be quick) and decide based on the results if speed feature should be disabled for low-res and/or screen content.
> > 
> > 
> > Test results of A3-B1 are updated in the description (the anchor's UPSNR of Vertical_Bayshore_270x480_2997 is not monotonic so the U result of class A5 is not available). B2 sequences are also tested but the trade off did not meet the bar so I have disabled it for screen content. B2 results are as follows for your reference.
> > ```
> > +--------------------------+---------+---------+---------+---------+----------+----------+
> > | Summary                  | Y       | U       | V       | YUV     | Enc-time | Dec-time |
> > +--------------------------+---------+---------+---------+---------+----------+----------+
> > | Class B2_SCC             | 0.66%   | 1.04%   | 1.03%   | 0.71%   | 97.2%    | 101%     |
> > +--------------------------+---------+---------+---------+---------+----------+----------+
> > ```
> 
> Thanks for the update!
> 
> Based on the results, we should also disable the speed feature for <= 270p. I have done something similar for another speed feature: https://github.com/AOMediaCodec/avm/blob/av2-enc/av2/encoder/speed_features.c#L1053

I have disabled it also for <= 270p cases in the latest version. Thanks for pointing this out.

av2/encoder/speed_features.c
motongxu: The default ratio (even not given from command line) is 8, so the speed feature will never work if command line cfg is prioritized. I think the user may turn off this speed feature if they would like to use command line value instead. Would it make sense?

urvangjoshi: I'm asking about the case when user asks for ratio 2 for example.  Currently, that will be ignored and reset to 4, right?

So, my suggested fix is to consider both:  user provided value and the speed feature based value and select the more restrictive of the two here.

yeqing-wu:
> > > > > > The loss on A2 is on higher side -- but ratio maybe acceptable.
> > > > > > But can you share results for other sets? I'm worried that loss maybe even higher for lower resolutions and also for B1 / B2.
> > > > > 
> > > > > 
> > > > > We didn't test class B1 and B2. We can turn off this speed feature for screen content by checking !cpi->is_screen_content_type. What do you think? @urvangjoshi
> > > > 
> > > > 
> > > > We should decide based on data. Perhaps, run B2 set and A5 set RA 33 frames at least (should be quick) and decide based on the results if speed feature should be disabled for low-res and/or screen content.
> > > 
> > > 
> > > Test results of A3-B1 are updated in the description (the anchor's UPSNR of Vertical_Bayshore_270x480_2997 is not monotonic so the U result of class A5 is not available). B2 sequences are also tested but the trade off did not meet the bar so I have disabled it for screen content. B2 results are as follows for your reference.
> > > ```
> > > +--------------------------+---------+---------+---------+---------+----------+----------+
> > > | Summary                  | Y       | U       | V       | YUV     | Enc-time | Dec-time |
> > > +--------------------------+---------+---------+---------+---------+----------+----------+
> > > | Class B2_SCC             | 0.66%   | 1.04%   | 1.03%   | 0.71%   | 97.2%    | 101%     |
> > > +--------------------------+---------+---------+---------+---------+----------+----------+
> > > ```
> > 
> > 
> > Thanks for the update!
> > Based on the results, we should also disable the speed feature for <= 270p. I have done something similar for another speed feature: https://github.com/AOMediaCodec/avm/blob/av2-enc/av2/encoder/speed_features.c#L1053
> 
> I have disabled it also for <= 270p cases in the latest version. Thanks for pointing this out.

Based on this B2 results, should we also disable it for B2?

av2/encoder/speed_features.c
motongxu: I understand your concern. However, the current default value of `max-pb-aspect-ratio` is 8, so the final value will always be 8 when using `AVMMAX(cpi->common.seq_params.max_pb_aspect_ratio_log2_m1, new_max_pb_aspect_ratio_log2_m1)`. In this case, this speed feature is only valid when command line value is smaller than 4.

urvangjoshi: Does AVMMIN work then?

motongxu:
> > > > > > > The loss on A2 is on higher side -- but ratio maybe acceptable.
> > > > > > > But can you share results for other sets? I'm worried that loss maybe even higher for lower resolutions and also for B1 / B2.
> > > > > > 
> > > > > > 
> > > > > > We didn't test class B1 and B2. We can turn off this speed feature for screen content by checking !cpi->is_screen_content_type. What do you think? @urvangjoshi
> > > > > 
> > > > > 
> > > > > We should decide based on data. Perhaps, run B2 set and A5 set RA 33 frames at least (should be quick) and decide based on the results if speed feature should be disabled for low-res and/or screen content.
> > > > 
> > > > 
> > > > Test results of A3-B1 are updated in the description (the anchor's UPSNR of Vertical_Bayshore_270x480_2997 is not monotonic so the U result of class A5 is not available). B2 sequences are also tested but the trade off did not meet the bar so I have disabled it for screen content. B2 results are as follows for your reference.
> > > > ```
> > > > +--------------------------+---------+---------+---------+---------+----------+----------+
> > > > | Summary                  | Y       | U       | V       | YUV     | Enc-time | Dec-time |
> > > > +--------------------------+---------+---------+---------+---------+----------+----------+
> > > > | Class B2_SCC             | 0.66%   | 1.04%   | 1.03%   | 0.71%   | 97.2%    | 101%     |
> > > > +--------------------------+---------+---------+---------+---------+----------+----------+
> > > > ```
> > > 
> > > 
> > > Thanks for the update!
> > > Based on the results, we should also disable the speed feature for <= 270p. I have done something similar for another speed feature: https://github.com/AOMediaCodec/avm/blob/av2-enc/av2/encoder/speed_features.c#L1053
> > 
> > 
> > I have disabled it also for <= 270p cases in the latest version. Thanks for pointing this out.
> 
> Based on this B2 results, should we also disable it for B2?

This speed feature is already disabled at SCC type due to its huge loss.

av2/encoder/speed_features.c
motongxu: I just added the condition of choosing a less strict value in the new version. Thank you.

urvangjoshi: Looks good now, thanks.

urvangjoshi:
> Based on this B2 results, should we also disable it for B2?
> 
> This speed feature is already disabled at SCC type due to its huge loss.

I think the PR description is making it confusing.
Could you update results for A5 and B2 there (to say 0.0 BDRate and runtime)?

Also, pls rebase locally and force-push to fix merge conflicts.

motongxu:
> > Based on this B2 results, should we also disable it for B2?
> > This speed feature is already disabled at SCC type due to its huge loss.
> 
> I think the PR description is making it confusing. Could you update results for A5 and B2 there (to say 0.0 BDRate and runtime)?
> 
> Also, pls rebase locally and force-push to fix merge conflicts.

The description is updated and conflicts are fixed.

PR #5248, opened 2026-08-12
Add rule-based partition pruning at cpu-used=1

YixinDu1:
Introduces rule-based partition pruning via calibrated RD-cost lookup tables (LUTs). Four pruning rules are applied at cpu-used=1:

- 4-way (HORZ-4A/4B, VERT-4A/4B): disabled unconditionally for eligible intra block sizes.
- 3-way (HORZ-3, VERT-3): pruned when NONE is the current best.
- RECT (HORZ, VERT): pruned when rd_cost_none is below a per- (bsize, bitdepth, qindex) LUT threshold.
- SPLIT: pruned when rd_cost_none is below a per-(bsize, qp-band) threshold (10-bit only).

Anchor: commit 65a8c61a
Performance, cpu-used=1, CTC AI+RA (10 frames AI / 33 frames RA):

AI:
```
+------------+-------+-------+-------+-------+------+
| Class      |     Y |    Cb |    Cr |  wAvg | Enc% |
+------------+-------+-------+-------+-------+------+
| A1         | +0.15 | +0.50 | -0.02 | +0.16 |   83 |
| A2         | +0.28 | +0.56 | +0.51 | +0.31 |   77 |
+------------+-------+-------+-------+-------+------+
| Avg w/o B2 | +0.26 | +0.75 | +0.52 | +0.30 |   75 |
+------------+-------+-------+-------+-------+------+
```
RA:
```
+------------+-------+-------+-------+-------+------+
| Class      |     Y |    Cb |    Cr |  wAvg | Enc% |
+------------+-------+-------+-------+-------+------+
| A1         | +0.09 | +0.59 | +0.55 | +0.14 |   95 |
| A2         | +0.09 | +0.42 | +0.62 | +0.12 |   92 |
+------------+-------+-------+-------+-------+------+
| Avg w/o B2 | +0.09 | +0.31 | +0.72 | +0.12 |   92 |
+------------+-------+-------+-------+-------+------+
```
STATS_CHANGED


YixinDu1:
CC: @YunfeiZheng

urvangjoshi:
@YixinDu1 :  Some general comments:
(1) The loss seems to be high, especially for intra.   I'd suggest considering such changes for faster speed settings instead.
(2) Do you have a breakdown of losses from each change?
(3)  I have a feeling that disabling 4A/4B itself should have close to 0.3% loss for AI.

urvangjoshi (changes requested):
See my other comment

YixinDu1:
> @YixinDu1 : Some general comments: (1) The loss seems to be high, especially for intra. I'd suggest considering such changes for faster speed settings instead. (2) Do you have a breakdown of losses from each change? (3) I have a feeling that disabling 4A/4B itself should have close to 0.3% loss for AI.

@urvangjoshi The idea here is to not disable certain partition types unconditionally, but use some learned rules from encoder to guide it. For intra class A1, we have reached a ratio of 17/0.16=106x, does this not look good you? What do @leolzhao and @yunqingwang1 think about this trade off for intra? How about RA?

urvangjoshi:
> > @YixinDu1 : Some general comments: (1) The loss seems to be high, especially for intra. I'd suggest considering such changes for faster speed settings instead. (2) Do you have a breakdown of losses from each change? (3) I have a feeling that disabling 4A/4B itself should have close to 0.3% loss for AI.
> 
> @urvangjoshi The idea here is to not disable certain partition types unconditionally, but use some learned rules from encoder to guide it. For intra, we have reached a ratio of 25/0.3=83x, does this not look good you? What do @leolzhao and @yunqingwang1 think about this trade off for intra? How about RA?

@YixinDu1 : My concerns are:
(1) Absolute loss for intra is very high -- even if the ratio may look ok.
(2) I'm concerned that the loss from this change is as high as simply disabling 4a/4b partitions blindly. 

That's why I'm curious what the breakdown is from different changes -- to potentially find a better solution / smaller loss operating point.

YixinDu1:
> > > @YixinDu1 : Some general comments: (1) The loss seems to be high, especially for intra. I'd suggest considering such changes for faster speed settings instead. (2) Do you have a breakdown of losses from each change? (3) I have a feeling that disabling 4A/4B itself should have close to 0.3% loss for AI.
> > 
> > 
> > @urvangjoshi The idea here is to not disable certain partition types unconditionally, but use some learned rules from encoder to guide it. For intra, we have reached a ratio of 25/0.3=83x, does this not look good you? What do @leolzhao and @yunqingwang1 think about this trade off for intra? How about RA?
> 
> @YixinDu1 : My concerns are: (1) Absolute loss for intra is very high -- even if the ratio may look ok. (2) I'm concerned that the loss from this change is as high as simply disabling 4a/4b partitions blindly.
> 
> That's why I'm curious what the breakdown is from different changes -- to potentially find a better solution / smaller loss operating point.

Thanks for the clarification. How about trade-off for RA? Does that look ok to you?

yeqing-wu:
Has anyone tested 4a/4b partitions? I recall @leolzhao mentioning this earlier. Let’s first have the results of the 4a/4b partitions, and then we can decide on the next steps.

yunqingwang1:
We recently tested "part_sf.disable_uneven_4way_partitions = 1" (vs Commit 587f50e) for speed 1. The average enc time reduction for a1 & a2 set is 8.7% and the tradeoff meets speed 1 bar (A1: 0.09%; A2: 0.2%).

YixinDu1:
Hi @urvangjoshi, please find the results of disabling 4-way for cpu1 attached. 
image

PR #5253, opened 2026-08-12
Two-pass superblock partition search

yeqing-wu:
Speed up the partition search by splitting it into two passes over each superblock.

The first pass (the "dry" pass) works out a rough partition shape. It runs mode search with a reduced set of tools, because it only has to rank shapes against each other, not produce the final coding decision. It stops at 16x16.

The second pass (the "wet" pass) codes the superblock for real, with the full set of tools. It keeps the first pass's shape for blocks of 32x32 and larger and searches the smaller blocks again, since that is where the first pass is least reliable. Large blocks the first pass left unsplit are also searched again, because a reduced tool set tends to under-split them, and trusting that would leave the block over-merged with no way to recover. That re-search is limited to blocks up to 128 and stops at 8x8, which keeps most of the time saved.

STATS_CHANGED

Anchor: commit 5b32726
Speed 3 (cpu-used=3): FG16 CTC (33 frames, class A1 and A2, RA)
Speed 2 (cpu-used=2): FG16 CTC (33 frames, class A1 and A2, RA)

```
    1) Speed 4 (cpu-used=4)
      | A1         | -0.09 | -2.32 | -2.75 | -0.35 |   86 |   86 |
      | A2         |  0.78 |  0.86 |  1.13 |  0.80 |   73 |   95 |
      | Avg w/o B2 |  0.52 | -0.08 | -0.02 |  0.46 |   77 |   93 |
    
    2) Speed 3 (cpu-used=3)
      | A1         | -0.03 | -2.27 | -2.79 | -0.29 |   93 |   88 |
      | A2         |  0.81 |  0.98 |  1.82 |  0.87 |   73 |   96 |
      | Avg w/o B2 |  0.56 |  0.02 |  0.45 |  0.53 |   79 |   94 |
    
    3) Speed 2 (cpu-used=2): Not satisfied: Score = 24.7
      | A1         |  0.39 | -1.90 | -2.76 |  0.11 |   94 |   89 |
      | A2         |  0.94 |  2.07 |  2.25 |  1.05 |   76 |   96 |
      | Avg w/o B2 |  0.77 |  0.89 |  0.77 |  0.77 |   81 |   94 |
```

Thus, it is only enabled for speed >= 3 in the current stage.

### DIFF
--- a/av2/encoder/encodeframe.c
+++ b/av2/encoder/encodeframe.c
@@ -594,12 +594,18 @@ static AVM_INLINE void perform_two_partition_passes(
                              SB_WET_PASS, NULL);
 }
 
-/*!\brief Set all tree nodes <= min_bsize to PARTITION_INVALID.
+/*!\brief Mark the small nodes of the dry-pass tree as "search again".
  *
  * \ingroup partition_search
+ *
+ * The wet pass trusts the dry-pass shape for large blocks and searches the
+ * small ones again. This walks the dry-pass tree and clears the shape of every
+ * node smaller than min_bsize so the wet pass is free to choose its own.
  */
 static AVM_INLINE void set_min_none_to_invalid(PARTITION_TREE *part_tree,
-                                               BLOCK_SIZE min_bsize) {
+                                               BLOCK_SIZE min_bsize,
+                                               bool allow_none_resplit) {
+  if (!part_tree) return;
   const BLOCK_SIZE bsize = part_tree->bsize;
   const PARTITION_TYPE part_type = part_tree->partition;
   if (!is_bsize_geq(bsize, min_bsize)) {
@@ -608,7 +614,19 @@ static AVM_INLINE void set_min_none_to_invalid(PARTITION_TREE *part_tree,
       av2_free_ptree_recursive(part_tree->sub_tree[idx]);
       part_tree->sub_tree[idx] = NULL;
     }
+    return;
+  }
 
+  // Large blocks the dry pass left unsplit: re-split them in the wet pass
+  // (only on the fast level, guarded by allow_none_resplit). The conservative
+  // level keeps its pre-PR semantics.
+  if (allow_none_resplit && part_type == PARTITION_NONE) {
+    // Only do this for blocks up to 128 px. Bigger unsplit blocks are usually
+    // genuinely flat, so searching them again costs a lot of time for little
+    // gain.
+    if (AVMMAX(block_size_wide[bsize], block_size_high[bsize]) <= 128) {
+      part_tree->partition = PARTITION_INVALID;
+    }
     return;
   }
 
@@ -630,7 +648,8 @@ static AVM_INLINE void set_min_none_to_invalid(PARTITION_TREE *part_tree,
   }
 
   for (int idx = 0; idx < num_subtrees; idx++) {
-    set_min_none_to_invalid(part_tree->sub_tree[idx], min_bsize);
+    set_min_none_to_invalid(part_tree->sub_tree[idx], min_bsize,
+                            allow_none_resplit);
   }
 }
 
@@ -644,7 +663,8 @@ static AVM_INLINE void set_min_none_to_invalid(PARTITION_TREE *part_tree,
  */
 static AVM_INLINE void perform_two_pass_partition_search(
     AV2_COMP *cpi, ThreadData *td, TileDataEnc *tile_data, TokenExtra **tp,
-    TokenExtra **tp_chroma, const int mi_row, const int mi_col) {
+    TokenExtra **tp_chroma, const int mi_row, const int mi_col,
+    bool fast_two_pass) {
   SIMPLE_MOTION_DATA_TREE *const sms_root = td->sms_root;
   AV2_COMMON *const cm = &cpi->common;
   MACROBLOCK *const x = &td->mb;
@@ -653,17 +673,20 @@ static AVM_INLINE void perform_two_pass_partition_search(
   const BLOCK_SIZE sb_size = cm->sb_size;
   assert(!frame_is_intra_only(cm));
 
-  // First pass to estimate  partition structures
+  // First pass to estimate partition structures
   SB_FIRST_PASS_STATS sb_fp_stats;
   av2_backup_sb_state(&sb_fp_stats, cpi, td, tile_data, mi_row, mi_col);
-  const BLOCK_SIZE fp_min_bsize = BLOCK_16X16;
-  x->sb_enc.min_partition_size = fp_min_bsize;
+  // The dry pass does not go below 16x16: it only needs a rough shape, and it
+  // scores blocks with a reduced set of tools.
+  x->sb_enc.min_partition_size = BLOCK_16X16;
   perform_one_partition_pass(cpi, td, tile_data, tp, tp_chroma, mi_row, mi_col,
                              SB_DRY_PASS, NULL);
   PARTITION_TREE *part_ref = xd->sbi->ptree_root[0];
   // Set this to NULL otherwise part_ref will get freed in the second pass.
   xd->sbi->ptree_root[0] = NULL;
-  set_min_none_to_invalid(part_ref, get_larger_sqr_bsize(fp_min_bsize));
+  // Trust the dry-pass shape for >=32x32 blocks; re-search smaller ones.
+  set_min_none_to_invalid(part_ref, get_larger_sqr_bsize(BLOCK_16X16),
+                          fast_two_pass);
 
   // Second pass
   RD_STATS dummy_rdc;
@@ -850,9 +873,13 @@ static AVM_INLINE void encode_rd_sb(AV2_COMP *cpi, ThreadData *td,
                                    mi_col);
     } else if (!frame_is_intra_only(cm) &&
                bru_is_sb_active(cm, mi_col, mi_row) &&
-               sf->part_sf.two_pass_partition_search) {
-      perform_two_pass_partition_search(cpi, td, tile_data, tp, tp_chroma,
-                                        mi_row, mi_col);
+               av2_two_pass_part_enabled(&sf->part_sf)) {
+      // TODO(Yeqing): Add an SB-level heuristic to skip the second pass on
+      // SBs where a single pass is good enough, to reduce encoding time.
+      // Only the fast level opts into the extra re-split of unsplit blocks.
+      perform_two_pass_partition_search(
+          cpi, td, tile_data, tp, tp_chroma, mi_row, mi_col,
+          av2_two_pass_part_is_fast(&sf->part_sf));
     } else {
       perform_one_partition_pass(cpi, td, tile_data, tp, tp_chroma, mi_row,
                                  mi_col, SB_SINGLE_PASS, NULL);
--- a/av2/encoder/encoder_utils.c
+++ b/av2/encoder/encoder_utils.c
@@ -533,7 +533,6 @@ void av2_scale_references(AV2_COMP *cpi, const InterpFilter filter,
 }
 
 BLOCK_SIZE av2_select_sb_size(const AV2_COMP *const cpi) {
-  const AV2_COMMON *const cm = &cpi->common;
   const AV2EncoderConfig *const oxcf = &cpi->oxcf;
 
   if (oxcf->tool_cfg.superblock_size == AVM_SUPERBLOCK_SIZE_64X64)
@@ -545,28 +544,40 @@ BLOCK_SIZE av2_select_sb_size(const AV2_COMP *const cpi) {
 
   assert(oxcf->tool_cfg.superblock_size == AVM_SUPERBLOCK_SIZE_DYNAMIC);
 
+  BLOCK_SIZE chosen;
   if (oxcf->resize_cfg.resize_mode != RESIZE_NONE) {
     // Use the configured size (top resolution) for spatial layers or
     // on resize.
-    return AVMMIN(oxcf->frm_dim_cfg.width, oxcf->frm_dim_cfg.height) >= 720
-               ? BLOCK_256X256
-           : AVMMIN(oxcf->frm_dim_cfg.width, oxcf->frm_dim_cfg.height) > 480
-               ? BLOCK_128X128
-               : BLOCK_64X64;
+    chosen = AVMMIN(oxcf->frm_dim_cfg.width, oxcf->frm_dim_cfg.height) >= 720
+                 ? BLOCK_256X256
+             : AVMMIN(oxcf->frm_dim_cfg.width, oxcf->frm_dim_cfg.height) > 480
+                 ? BLOCK_128X128
+                 : BLOCK_64X64;
+  } else if (oxcf->speed > 1 && !av2_wants_two_pass_partition(oxcf)) {
+    // TODO(any): Possibly could improve this with a heuristic.
+    // When resize is on, 'cm->width / height' can change between calls, so we
+    // don't apply this heuristic there. Things break if superblock size
+    // changes between the first pass and second pass encoding, which is why
+    // this heuristic is not configured as a speed-feature. Two-pass partition
+    // search requires a large superblock, so the small-SB heuristic is
+    // guarded on the same predicate.
+    const AV2_COMMON *const cm = &cpi->common;
+    chosen = AVMMIN(cm->width, cm->height) > 480 ? BLOCK_128X128 : BLOCK_64X64;
+  } else {
+    chosen = AVMMIN(oxcf->frm_dim_cfg.width, oxcf->frm_dim_cfg.height) >= 720
+                 ? BLOCK_256X256
+                 : BLOCK_128X128;
   }
 
-  // TODO(any): Possibly could improve this with a heuristic.
-  // When resize is on, 'cm->width / height' can change between
-  // calls, so we don't apply this heuristic there.
-  // Things break if superblock size changes between the first pass and second
-  // pass encoding, which is why this heuristic is not configured as a
-  // speed-feature.
-  if (oxcf->resize_cfg.resize_mode == RESIZE_NONE && oxcf->speed > 1) {
-    return AVMMIN(cm->width, cm->height) > 480 ? BLOCK_128X128 : BLOCK_64X64;
+  // Never choose an SB larger than the shortest side of the frame. This
+  // avoids a known issue where two-pass partition on multi-layer content
+  // produces a malformed OBU when the SB spans a small frame.
+  const int short_side =
+      AVMMIN(oxcf->frm_dim_cfg.width, oxcf->frm_dim_cfg.height);
+  while (chosen != BLOCK_64X64 && block_size_wide[chosen] > short_side) {
+    chosen = (chosen == BLOCK_256X256) ? BLOCK_128X128 : BLOCK_64X64;
   }
-  return AVMMIN(oxcf->frm_dim_cfg.width, oxcf->frm_dim_cfg.height) >= 720
-             ? BLOCK_256X256
-             : BLOCK_128X128;
+  return chosen;
 }
 
 void reallocate_sb_size_dependent_buffers(AV2_COMP *cpi) {
--- a/av2/encoder/encoder_utils.h
+++ b/av2/encoder/encoder_utils.h
@@ -959,6 +959,17 @@ void av2_scale_references(AV2_COMP *cpi, const InterpFilter filter,
 
 void av2_setup_frame(AV2_COMP *cpi);
 
+// Config-time part of the TWO_PASS_PART_FAST predicate: GOOD mode at speed >= 3
+// on configs that produce inter frames. Kept config-only (no cm state) because
+// av2_select_sb_size() also keys off it, and the superblock size must not
+// change between the two passes.
+// TODO(Yeqing): Extend to intra/key frames and drop the all_intra guard.
+static INLINE bool av2_wants_two_pass_partition(const AV2EncoderConfig *oxcf) {
+  const int all_intra =
+      oxcf->kf_cfg.key_freq_max == 0 && oxcf->kf_cfg.key_freq_min == 0;
+  return oxcf->mode == GOOD && oxcf->speed >= 3 && !all_intra;
+}
+
 BLOCK_SIZE av2_select_sb_size(const AV2_COMP *const cpi);
 void set_ard_active_map(AV2_COMP *cpi);
 
--- a/av2/encoder/interp_search.h
+++ b/av2/encoder/interp_search.h
@@ -46,6 +46,51 @@ typedef struct InterModeRdPair {
 #define NUM_USE_AMVD_OPTIONS 2
 /*!\endcond */
 
+/*! \brief Dry-pass caps for the two-pass partition search. */
+typedef struct {
+  /*! \brief Highest rank of inter reference searched (0-based). */
+  int ref_rank_cap;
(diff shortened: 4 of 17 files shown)


av2/encoder/speed_features.h
```diff
@@ -401,6 +401,11 @@ typedef struct PARTITION_SPEED_FEATURES {
 
   int two_pass_partition_search;
 
+  // When set, run the two-pass partition search on every superblock of a
```
urvangjoshi: A number of high level questions first:

(1) Not sure if I understood the comment correctly.   I believe `two_pass_partition_search` is also a 2-pass search at superblock level, correct? 

(2) Can you explain what are the differences between the two and when is one turned or vs the other? 

(3) Any reason not to reuse the same speed feature and modify it?

(4) I believe `two_pass_partition_search` is only enabled at high resolutions.   Is that the same for `two_pass_partition_search_sb_level` as well?  Or is it enabled for more (lower) resolutions too?

yeqing-wu: Hi @urvangjoshi, please see my reply inlined.

> (1) Not sure if I understood the comment correctly. I believe two_pass_partition_search is also a 2-pass search at superblock level, correct?

Yeqing: The two_pass_partition_search feature is currently only enabled for high-resolution and low QPs (almost only QP=110). It is determined at the frame level. If enabling it for high QP, there is huge encoding time increase. However, two_pass_partition_search_sb_level allows for toggling this feature at the superblock level. As my TODO comments, I am working on a simple ML or heuristic algorithm to implement this feature. Additionally, two_pass_partition_search_sb_level can be enabled for all QPs and resolutions. 

> (2) Can you explain what are the differences between the two and when is one turned or vs the other?

Yeqing: For speeds 0 to 2, since this PR doesn’t achieve a good tradeoff (I believe speed 2 is close to the ratio, and I’ll do my best to make it happen), thus the code turns off two_pass_partition_search_sb_level and turned on two_pass_partition_search (as mentioned in (1), only for high resolution and low QPs) for speed 0 to 2. For speeds 3 and above, I’ve only enabled two_pass_partition_search_sb_level. So, two_pass_partition_search and two_pass_partition_search_sb_level are exclusive. In the future, I’ll try to merge their functionalities into one, but so far, I haven’t been able to achieve a good tradeoff, so I’ve kept both of them for different purposes. 

> (3) Any reason not to reuse the same speed feature and modify it?

Yeqing: see my answer in (2)

> (4) I believe two_pass_partition_search is only enabled at high resolutions. Is that the same for two_pass_partition_search_sb_level as well? Or is it enabled for more (lower) resolutions too?

Yeqing: see my answer in (1), two_pass_partition_search is only enabled for 1080p and above, and only for very low QP (almost only QP=110). two_pass_partition_search_sb_level is enabled for all resolutions and all QPs.

Thanks
Yeqing

av2/encoder/encoder_utils.h
```diff
@@ -959,6 +959,15 @@ void av2_scale_references(AV2_COMP *cpi, const InterpFilter filter,
 
 void av2_setup_frame(AV2_COMP *cpi);
 
+// Config-time query for the sb-level two-pass partition regime. Enabled in
+// GOOD mode at speed >= 3 on configs that produce inter frames.
+// TODO(Yeqing): Extend to intra/key frames and drop the all_intra guard.
+static INLINE bool av2_wants_two_pass_partition(const AV2EncoderConfig *oxcf) {
+  const int all_intra =
+      oxcf->kf_cfg.key_freq_max == 0 && oxcf->kf_cfg.key_freq_min == 0;
+  return oxcf->mode == GOOD && oxcf->speed >= 3 && !all_intra;
```
leolzhao, line 970: why is it restricted to Good mode? Can we turn it on by default for RT as well? If it doesn't have good performance for RTC, we can turn it off later?

av2/encoder/speed_features.h
leolzhao: One follow-up question: what's the difference between `two_pass_partition_search` and `two_pass_partition_search_sb_level` in terms of algorithm?

av2/encoder/encodeframe.c
```diff
@@ -608,7 +614,19 @@ static AVM_INLINE void set_min_none_to_invalid(PARTITION_TREE *part_tree,
...
+  }
 
+  // Large blocks the dry pass left unsplit: re-split them in the wet pass
+  // (only on the SB-level path, guarded by allow_none_resplit). The frame-
+  // level path keeps its pre-PR semantics.
+  if (allow_none_resplit && part_type == PARTITION_NONE) {
+    // Only do this for blocks up to 128 px. Bigger unsplit blocks are usually
+    // genuinely flat, so searching them again costs a lot of time for little
+    // gain.
+    if (AVMMAX(block_size_wide[bsize], block_size_high[bsize]) <= 128) {
```
leolzhao, line 627: Does this mean that wet pass doesn't trust the decided partition_none type from dry pass for block width and height <= 128?

av2/encoder/speed_features.h
yeqing-wu: 1. two_pass_partition_search uses the full inter/intra/transform tools set in both the first pass and the second pass, thus it would increase encoding time a lot if we enable it for all resolutions and all QPs. For two_pass_partition_search_sb_level, we only use very basic tools in the dry pass to help decide the partition decision and avoid increasing the encoding time;

2. For two_pass_partition_search, all >= 32x32, it would directly use the partition decision from the first pass without the chance to tune it. This is not good for the performance.

3. Meanwhile, for two_pass_partition_search_sb_level, we will allow it to be turned on and off at the SB level, instead of the frame level. This would achieve better encoding time. As indicated in the code, this is TODO item, I am implementing this one.

yeqing-wu: As I mentioned earlier, I will do my best to merge these two concepts into a single entity in the future. However, it could potentially affect the performance of speed 0, and it may require lots of effort to make it work effectively.

av2/encoder/encodeframe.c
yeqing-wu: Yes, we allow it has some room to tune it. This can help reduce the loss. Meanwhile, as you can see, only if it is PARTITION_NONE, we allow to tune it. Thus we still keep the partition decision from top to this level

av2/encoder/encoder_utils.h
yeqing-wu: I did not test this one for the RT case, for RT, I believe lots of tools are turned off, maybe we do not need this two pass partition. I think we can turn it on for testing in the future once RT code is stable. I saw we are actively working on RT mode, not very stable.

yeqing-wu:
@urvangjoshi and @leolzhao,

I have unified part_sf.two_pass_partition_search and part_sf.two_pass_partition_search_sb_level with a single levelled speed feature part_sf.two_pass_partition_search with the enum type, I think this design is better and clean.
    
```
      TWO_PASS_PART_OFF          (0)
      TWO_PASS_PART_CONSERVATIVE (1)  full-tool dry pass, trust unsplit shapes
      TWO_PASS_PART_FAST         (2)  reduced-tool dry pass, re-split unsplit
```

Thanks
Yeqing

av2/encoder/partition_search.c
```diff
@@ -5445,10 +5264,29 @@ bool av2_rd_pick_partition(
   if (bsize == cm->sb_size)
     pc_tree->is_cfl_allowed_for_this_chroma = CFL_DISALLOWED_FOR_CHROMA;
 
+  int eff_max_recursion_depth = max_recursion_depth;
+  // Cap the wet-pass re-split at ~8x8 (short-side driven). Without a cap it
```
yunqingwang1, line 5451: Does this wet-pass re-split only happens when the first pass decision is NONE for this whole partition block? Also, if first partition decision is NONE, it could imply that this is a flat partition. Do you always need to go down to 8x8 or a smaller depth can be enough?

av2/encoder/rdopt.c
```diff
@@ -8872,6 +8941,43 @@ void av2_rd_pick_inter_mode_sb(struct AV2_COMP *cpi,
...
+      .compound_type_mask = (1u << COMPOUND_AVERAGE) | (1u << COMPOUND_WEDGE),
+      .inter_mode_mask = (1u << NEWMV) | (1u << NEW_NEWMV) | (1u << NEARMV) |
+                         (1u << NEAR_NEARMV) | (1u << WARPMV) |
+                         (1u << WARP_NEWMV),
+      .precisions_from_top = 3,
+      .intra_mpm_cap = 13,
+    };
+    // Keep the normal distortion measure; a cheaper one made the partition
+    // choices worse.
+    x->rd_model = FULL_TXFM_RD;
```
yunqingwang1, line 9015: So the transform search process is the same for both partition search passes?

yeqing-wu: I tried LOW_TXFM_RD, but it does not help. Meanwhile, I also tried do not run tx type and size search, this is not helpful. I think it impacts the RDO decision a lot. Thus I only gate for only allowing 2 fast rd candidates to pass to the full rd even with a limited very small basic tool set in the dry pass if it is TWO_PASS_PART_FAST (new introduced simplified two pass). Do you have any suggestion related to transform search in the dry pass for future improvements? I still have some ideas to further improve this TWO_PASS_PART_FAST.

```
  if (x->apply_dry_pass_shortcuts) {
    txfm_params->tx_mode_search_type = TX_MODE_LARGEST;
    txfm_params->tx_size_search_method = USE_LARGESTALL;
    txfm_params->use_default_intra_tx_type = 1;
    txfm_params->use_default_inter_tx_type = 1;
  }
```

av2/encoder/partition_search.c
yeqing-wu: Yes, your understand is correct. I tried to cap at 8, 16, 32, but cap at 8x8 is the best tradeoff. And the encoding time different is not obvious. We can design to have a different cap for different speed preset based on the encoding time saving and the loss in the future.

yunqingwang1: Thanks.

av2/encoder/rdopt.c
yunqingwang1: Thanks for the explanation! As you mentioned, using the largest tx size and DCT only for first pass was something we did before in AV1. But not sure about the tradeoff for AV2.

yeqing-wu:
Hi All,

Since it has passed all tests, could you please help review the code? If there is no issue, could you please help approve it?

Thanks
Yeqing

urvangjoshi:
@yeqing-wu :  thanks for the updates.   My main overall concern is the mix of gains (on A1) and losses (A2 and lower) -- but the speed-up is indeed very good.

Let me carefully review today / Monday to see if there is some issue / something where we can improve etc.

yeqing-wu:
Thanks @urvangjoshi , there are still lots of things we can improve for two pass partitions. We can do that step by step. And we can have different tunes for different speeds.

yeqing-wu:
Hi All,

Any comments on this PR? I'm currently working on per-QP and per-speed-preset configuration for two-pass partition which can achieve much better tradeoff, and in parallel I'm improving the two-pass partition structure to recover some of the coding gain (Already observed -0.13% gain with minor encoding time saving). Both of those build on this change, so I'd like to get it merged to continue with the next steps.

Thanks
Yeqing

av2/encoder/encoder_utils.c
```diff
@@ -545,28 +544,40 @@ BLOCK_SIZE av2_select_sb_size(const AV2_COMP *const cpi) {
...
-               ? BLOCK_256X256
-           : AVMMIN(oxcf->frm_dim_cfg.width, oxcf->frm_dim_cfg.height) > 480
-               ? BLOCK_128X128
-               : BLOCK_64X64;
+    chosen = AVMMIN(oxcf->frm_dim_cfg.width, oxcf->frm_dim_cfg.height) >= 720
+                 ? BLOCK_256X256
+             : AVMMIN(oxcf->frm_dim_cfg.width, oxcf->frm_dim_cfg.height) > 480
+                 ? BLOCK_128X128
+                 : BLOCK_64X64;
+  } else if (oxcf->speed > 1 && !av2_wants_two_pass_partition(oxcf)) {
```
urvangjoshi, line 556: Can you explain what exactly you are doing by adding the condition `!av2_wants_two_pass_partition()` here?  

Are we allowing 256 size superblock at faster speeds now, and we weren't allowing it before? 

If so, have you checked the effect of this change by itself?

av2/encoder/encodeframe.c
leolzhao: Thanks for the explanation.

av2/encoder/interp_search.h
```diff
@@ -46,6 +46,51 @@ typedef struct InterModeRdPair {
 #define NUM_USE_AMVD_OPTIONS 2
 /*!\endcond */
 
+/*! \brief Dry-pass caps for the two-pass partition search. */
+typedef struct {
```
urvangjoshi, line 50: This struct should go to something like `partition_search.h`  -- as it's unrelated to interpolation filter search.

av2/encoder/partition_search.c
```diff
@@ -5434,10 +5447,29 @@ bool av2_rd_pick_partition(
   if (bsize == cm->sb_size)
     pc_tree->is_cfl_allowed_for_this_chroma = CFL_DISALLOWED_FOR_CHROMA;
 
+  int eff_max_recursion_depth = max_recursion_depth;
```
urvangjoshi, line 5450: Is this code block trying to have min coding block size of 8X8?  If so, we can probably just use `x->sb_enc.min_partition_size`, right? 

Or is the logic different somehow? 

I'm worried that you are assuming half size at each level, which may not be true depending on which partition type is being used at each level.

leolzhao:
This MR looks good to me. @yeqing-wu

yeqing-wu:
> This MR looks good to me. @yeqing-wu

Thanks very much @leolzhao for helping review the PR.

urvangjoshi (commented):
We have had significant PRs for speed >= 3 recently.  Could you rerun the tests to see the latest results with your PR?

If results same as before, I'm slightly leaning towards enabling this for 4K resolution only for now -- and enable for others later after we tune it more.

I'll let others chime in as well though, and we can make a collective decision.

Some comments from partial review.

av2/encoder/interp_search.h
yeqing-wu: This data structure has the similar functionality as HandleInterModeArgs, thus from my opinion, it should stay the same place as HandleInterModeArgs.

av2/encoder/encoder_utils.c
yeqing-wu: Two-pass partition is designed to deciding partition size at large block sizes, so 256×256 down to 128×128 is the size range where it can be effective. Simply enabling it gives a −0.83% BD-rate gain at a 14% encoding time cost. It was previously disabled for speed > 1.

yeqing-wu:
> We have had significant PRs for speed >= 3 recently. Could you rerun the tests to see the latest results with your PR?
> 
> If results same as before, I'm slightly leaning towards enabling this for 4K resolution only for now -- and enable for others later after we tune it more.
> 
> I'll let others chime in as well though, and we can make a collective decision.
> 
> Some comments from partial review.

image

This is the results from latest tests, including CCSO etc. speed up. (without including the improvement I listed below)

Is there a particular reason we're enabling this only for 4K? From what I can see, it also achieves a very good tradeoff at 2K, the ratio is higher than the minimum bar, should we also enable it for 2k? Please let me know if I'm missing something. 

On my side, I'm continuing to improve two-pass partition and have already achieved
image

av2/encoder/interp_search.h
urvangjoshi: Yeah, I see that now.  That seems fine then.

urvangjoshi:
> > We have had significant PRs for speed >= 3 recently. Could you rerun the tests to see the latest results with your PR?
> > If results same as before, I'm slightly leaning towards enabling this for 4K resolution only for now -- and enable for others later after we tune it more.
> > I'll let others chime in as well though, and we can make a collective decision.
> > Some comments from partial review.
> 
> image
> This is the results from latest tests, including CCSO etc. speed up. (without including the improvement I listed below)
> 

Can you update MR description with this? 
I was curious if there was a different result especially after  283b7ca0b7e8c57bd62d5054dd366a45ac318566  and 7f268de0b8223d87398595458c595da19c6f1c7c.

I feel that this can have a large impact on the results.

> Is there a particular reason we're enabling this only for 4K? From what I can see, it also achieves a very good tradeoff at 2K, the ratio is higher than the minimum bar, should we also enable it for 2k? Please let me know if I'm missing something.

My thinking was:
- For 4K, there's both gain and speed-up.  So, it's obvious that we should turn on this feature -- and that way, your code also gets merged. [Turning speed feature on/off for other resolutions is easy later on].
- For 2K and lower resolutions,  (a) The absolute loss is a bit on the higher side, but (b) the ratio is good.  So, I'd like @yunqingwang1 @leolzhao to chime in and decide what we want to do.

> 
> On my side, I'm continuing to improve two-pass partition and have already achieved image

Can you explain these results?  Which anchor are they based on:   this PR as anchor,  or the baseline commit of this PR as anchor? 
In other words,  what will be the combined result of the 2 PRs?

yeqing-wu:
> Can you update MR description with this?
> I was curious if there was a different result especially after  https://github.com/AOMediaCodec/avm/commit/283b7ca0b7e8c57bd62d5054dd366a45ac318566 and https://github.com/AOMediaCodec/avm/commit/7f268de0b8223d87398595458c595da19c6f1c7c.

The results I've attached are already based on last Friday's commit. I can run a test with today's TCQ PR as well.  But I'd like to raise a concern: should we generally avoid asking the proponents of earlier PRs to re-run tests against PRs submitted several days later? If we ask for re-runs every time an incoming PR might impact others, the process could become open-ended, since most PRs have some overlap and their bases can differ significantly.

Regarding the gain shown in the table below, it is measured on top of the two-pass partitioning in this PR (already rebased on last Friday's ToT). The change uses the winning MV from the dry (first) pass as the starting MV for the wet pass, giving the wet pass a better starting point.

image

As I recall, we agreed in the FG16 group meeting that we should avoid bundling significant code changes into a single PR. Since the two-pass partitioning strategy is still evolving, I'd suggest we keep these changes in separate PRs rather than merging them together. Please let me know your thoughts.

av2/encoder/partition_search.c
yeqing-wu: Sorry that I missed this comment.

I think x->sb_enc.min_partition_size is intended for the entire SB, correct? If we set this value, even the block that doesn’t execute during the dry pass would be capped in this block size specified in sb_enc.min_partition_size, right?

urvangjoshi: You are right.  So just using the min_partition_size won't work.

Can you explain again which blocks you are trying to add the depth restriction for? 
In other words, what is the condition ```   if (template_tree && template_tree->partition == PARTITION_INVALID &&
      AVMMIN(block_size_wide[bsize], block_size_high[bsize]) >= 32) ``` checking?

Just trying to think if there is a better condition / check to do what you are intending to do.

urvangjoshi:
> > Can you update MR description with this?
> > I was curious if there was a different result especially after  [283b7ca](https://github.com/AOMediaCodec/avm/commit/283b7ca0b7e8c57bd62d5054dd366a45ac318566) and [7f268de](https://github.com/AOMediaCodec/avm/commit/7f268de0b8223d87398595458c595da19c6f1c7c).
> 
> The results I've attached are already based on last Friday's commit. I can run a test with today's TCQ PR as well. But I'd like to raise a concern: should we generally avoid asking the proponents of earlier PRs to re-run tests against PRs submitted several days later? If we ask for re-runs every time an incoming PR might impact others, the process could become open-ended, since most PRs have some overlap and their bases can differ significantly.

@yeqing-wu :   That's a fair point and agree that we should not ask for newer results in general.

My intention here of asking for newer results, was in fact to help with results from this PR.
My hope/wish was that you see even better results now that some unhelpful / buggy speed features have been disabled at speed 3+.  (And it's possible that those were hurting your PR as well).

> 
> Regarding the gain shown in the table below, it is measured on top of the two-pass partitioning in this PR (already rebased on last Friday's ToT). The change uses the winning MV from the dry (first) pass as the starting MV for the wet pass, giving the wet pass a better starting point.
> 
> image
> As I recall, we agreed in the FG16 group meeting that we should avoid bundling significant code changes into a single PR. Since the two-pass partitioning strategy is still evolving, I'd suggest we keep these changes in separate PRs rather than merging them together. Please let me know your thoughts.

To clarify:
- Yes,  MRs should remain separate
- The reason for clarifying the combined result was to see what to expect in the upcoming PR.

Now, IIUC,  the combined result from 2 PRs for A2 set will be 0.62% loss with 73% enc-time  in RA. Is that correct?

If so, I'm fine to go ahead with this if others are ok. @yunqingwang1 @leolzhao .

av2/encoder/encoder_utils.c
urvangjoshi: Will  −0.83% BD-rate gain at a 14% encoding time increase  be considered a good trade-off at speed 1?  (For FG agreed ratios)?

av2/encoder/partition_search.c
yeqing-wu: template_tree indicates that we have the partition information from the dry pass. Since we allow the re-opening of the partition during the wet pass (as explained in one of the previous comments), the recomputation of eff_max_recursion_depth only occurs when the re-opened blocks are of size >= 32×32. For these blocks (re-opened blocks of size at least 32×32) in the wet pass, we do not allow them to be partitioned < 8×8.

urvangjoshi: Got it.  Then the logic looks fine.

yeqing-wu:
> Now, IIUC, the combined result from 2 PRs for A2 set will be 0.62% loss with 73% enc-time in RA. Is that correct?

I haven’t tested their combination, but the gain (the MV from the dry pass as the starting MV for the wet pass) is within this fast two-pass partition, and saves very minor encoding time. Therefore, if we consider only the additive effect, the loss of A2 would decrease from 0.71% to (0.71% - 0.18%) = 0.53%. The encoding time savings should remain the same, even though the new changes would add 0.5% saving.

av2/encoder/encoder_utils.c
yeqing-wu: That is the result based on the first two commits of av2-enc (Cheng’s speed features adjustments). However, based on the latest av2-enc ToT, the encoding time may increase. I am currently running a test to verify this. Based on the limited data points, I checked the same QP and sequence, and the encoding time increased by about 4%. I think this is because the current av2-enc is significantly faster than before.

yeqing-wu:
If this two-pass partition fails to function as intended in the future, I would be more than willing to assist in removing the code responsible for it.

urvangjoshi:
> > Now, IIUC, the combined result from 2 PRs for A2 set will be 0.62% loss with 73% enc-time in RA. Is that correct?
> 
> I haven’t tested their combination, but the gain (the MV from the dry pass as the starting MV for the wet pass) is within this fast two-pass partition, and saves very minor encoding time. Therefore, if we consider only the additive effect, the loss of A2 would decrease from 0.71% to (0.71% - 0.18%) = 0.53%. The encoding time savings should remain the same, even though the new changes would add 0.5% saving.

Thanks, that sounds like a good trade-off indeed.

urvangjoshi (approved):

leolzhao (approved):

yunqingwang1 (approved):

PR #5258, opened 2026-08-12
Call memcpy() in avm_[highbd_]convolve_copy_*()

wantehchang:
```
Call memcpy() instead of memmove() in avm_convolve_copy_*() and
avm_highbd_convolve_copy_*(). avm_highbd_convolve_copy_avx2() and 
avm_highbd_convolve_copy_sse2() do not handle overlapping src and dst 
buffers when w is 256, so it is not necessary to call memmove().
```

### DIFF
--- a/avm_dsp/x86/avm_convolve_copy_avx2.c
+++ b/avm_dsp/x86/avm_convolve_copy_avx2.c
@@ -32,20 +32,20 @@ void avm_convolve_copy_avx2(const uint8_t *src, ptrdiff_t src_stride,
 
   if (w == 2) {
     do {
-      memmove(dst, src, 2 * sizeof(*src));
+      memcpy(dst, src, 2 * sizeof(*src));
       src += src_stride;
       dst += dst_stride;
-      memmove(dst, src, 2 * sizeof(*src));
+      memcpy(dst, src, 2 * sizeof(*src));
       src += src_stride;
       dst += dst_stride;
       h -= 2;
     } while (h);
   } else if (w == 4) {
     do {
-      memmove(dst, src, 4 * sizeof(*src));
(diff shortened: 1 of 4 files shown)


avm_dsp/x86/avm_convolve_copy_avx2.c
wantehchang, line 65: I believe the intrinsics code in this function (`avm_convolve_copy_avx2`) doesn't handle overlapping `src` and `dst` buffers. Please double check. Thanks!

jzern: In this block two source values are read and then two destination values are written. That should work.

jzern (approved):

jjustiss-apple (approved):

avm_dsp/x86/avm_convolve_copy_avx2.c
wantehchang: James: You are right and I was wrong. This copies each row of `src` and `dst` through a temporary buffer. So it is equivalent to `memmove()`.

wantehchang, line 302: I believe this code does not handle overlapping `src` and `dst` rows, so it is not equivalent to `memmove()`. Please double check.

Note: Everything else I checked is equivalent to `memmove()`. This is the only counter example I found in this file.

avm_dsp/x86/avm_convolve_copy_sse2.c
wantehchang, line 366: Similarly, this is not equivalent to `memmove()`, and this is also the only counter example I found in this file.

wantehchang:
Urvang: I just updated the commit message to reflect my latest findings.

PR #5272, opened 2026-08-17
Fast warp delta search for four and six param models in warp newmv

jadhuran:
Decouple warp parameter searches

Add cost based early termination

Enabled for speed levels four and above

Results

Anchor:  a341351f73129e48ee134579154c16b80496de05

Speed 4

```
+-------+-------+------+------+------+------+-------+
| Class |  Y%   |  U%  |  V%  | YUV% | ENC% | DEC%  |
+-------+-------+------+------+------+------+-------+
| A1    |  0.07 | 0.32 | 0.28 | 0.09 | 97.6 | 100.7 |
| A2    | -0.01 | 0.23 | 0.20 | 0.00 | 95.3 | 100.1 |
+-------+-------+------+------+------+------+-------+

```

### DIFF
--- a/av2/encoder/mcomp.c
+++ b/av2/encoder/mcomp.c
@@ -4921,7 +4921,7 @@ static void refine_translational_mv(
     WarpedMotionParams *best_wm_params, MV *best_mv,
     int warp_precision_idx_rate, const ModeCosts *mode_costs, int step_size,
     int max_coded_index, int num_neighbors, int num_mv_iterations,
-    uint64_t *best_rd) {
+    uint64_t *best_rd, const float cost_cmp_thr_speed) {
   static const MV neighbors[8] = { { 0, -1 }, { 1, 0 }, { 0, 1 },   { -1, 0 },
                                    { 1, -1 }, { 1, 1 }, { -1, -1 }, { -1, 1 } };
 
@@ -4970,6 +4970,12 @@ static void refine_translational_mv(
         if (this_rd < *best_rd) {
           best_idx = idx;
           *best_wm_params = *params;
+          if (cost_cmp_thr_speed != 0.f) {
+            if (this_rd * cost_cmp_thr_speed >= *best_rd) {
+              *best_rd = this_rd;
+              break;
+            }
+          }
           *best_rd = this_rd;
         }
       }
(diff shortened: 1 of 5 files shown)


av2/encoder/mcomp.c
```diff
@@ -5056,6 +5064,333 @@ static int get_valid_model_from_warp_stats_buffer(
...
+      // No need to change anything
+    }
+
+    if (center_best_so_far) {
+      break;
+    }
+  }
+}
+
+int av2_pick_warp_delta_fast(const AV2_COMMON *const cm, MACROBLOCKD *xd,
```
yunqingwang1: Is it possible to share the common code with av2_pick_warp_delta() instead of adding a new function here?

jadhuran: We can, but the code will be easy to read this way. There is a lot of conflict between these two functions.

av2/encoder/mcomp.c
```diff
@@ -4970,7 +4970,15 @@ static void refine_translational_mv(
...
-          *best_rd = this_rd;
+          if (cost_cmp_thr_speed != 0.f) {
+            if (this_rd * cost_cmp_thr_speed < *best_rd) {
+              *best_rd = this_rd;
+            } else {
+              *best_rd = this_rd;
+              break;
+            }
+          } else
+            *best_rd = this_rd;
```
yeqing-wu: in all cases, we assign *best_rd = this_rd;, thus is if/else branches redundant?

av2/encoder/mcomp.c
```diff
@@ -5056,6 +5064,333 @@ static int get_valid_model_from_warp_stats_buffer(
...
+  uint64_t rd;
+  int valid;
+  valid = av2_is_warp_model_reduced(params);
+  av2_get_shear_params(params, sf);
+  if (valid) {
+    av2_set_warp_translation(mi_row, mi_col, bsize, center_mv.as_mv, params);
+    rate = warp_precision_idx_rate +
+           av2_cost_model_param(mbmi, mode_costs, step_size, max_coded_index,
+                                &base_params);
+    sse = compute_motion_cost(xd, cm, ms_params, bsize, best_mv, can_refine_mv);
```
yeqing-wu, line 5084: the return type of compute_motion_cost() is unsigned int, thus should we use unsigned int for sse to avoid the casting?

av2/encoder/mcomp.c
yeqing-wu: I have the same comment as Yunqing, the duplicated code is huge between these two functions, it is better to avoid that, and let them share the same duplicated code.

av2/encoder/mcomp.c
```diff
@@ -5056,6 +5064,333 @@ static int get_valid_model_from_warp_stats_buffer(
...
+        *best_rd = inc_rd;
+      }
+    } else if (dec_rd < *best_rd) {
+      // Decreasing is best
+      if (dec_rd * cost_cmp_thr < *best_rd) center_best_so_far = 0;
+      *best_wm_params = dec_params;
+      *best_rd = dec_rd;
+    } else {
+      // Current is best
+      // No need to change anything
```
yeqing-wu: do we really need this else branch and the comments?

av2/encoder/mcomp.c
```diff
@@ -5056,6 +5064,333 @@ static int get_valid_model_from_warp_stats_buffer(
   return 0;
 }
 
+uint64_t get_warp_delta_param_cost(
+    const AV2_COMMON *const cm, MACROBLOCKD *xd, MB_MODE_INFO *mbmi,
+    const SUBPEL_MOTION_SEARCH_PARAMS *ms_params, const ModeCosts *mode_costs,
+    WarpedMotionParams base_params, WarpedMotionParams *params,
```
yeqing-wu: it is better to pass by pointer instead of pass by value for WarpedMotionParams base_params to avoid the cost of the copy of the data structure

yeqing-wu (commented):
Thanks. It is a good tradeoff. I have some comments, please help take a look

av2/encoder/mcomp.c
jadhuran: It is not redundant as there are two cases when we equate *best_rd = this_rd. I have now refactored this condition for better clarity.

av2/encoder/mcomp.c
jadhuran: addressed

av2/encoder/mcomp.c
jadhuran: Removed. Addressed

jadhuran:
> Thanks. It is a good tradeoff. I have some comments, please help take a look

I have addressed your comments

av2/encoder/mcomp.c
jadhuran: I have now added my changes in the existing av2_pick_warp_delta function.

yeqing-wu (approved):

PR #5284, opened 2026-08-18
rtc: Reuse pc_tree in non-RD partition search

jianj-g:
Pre-allocate and reuse the partition context tree (pc_root / PC_TREE) per thread in real-time non-RD partition search instead of allocating and freeing nodes for every superblock.

- Add pc_root to ThreadData and pre-allocate once per thread/frame.
- Add av2_reset_pmc to reset PICK_MODE_CONTEXT state on node reuse.
- In av2_nonrd_use_partition, lazily allocate child pc_tree nodes and reset existing PMC nodes across superblocks.

This change is bitexact for RTC with 2.5% speed up for QVGA and 1.5% speed up for VGA and HD on speed 6.

### DIFF
--- a/av2/encoder/context_tree.c
+++ b/av2/encoder/context_tree.c
@@ -200,6 +200,21 @@ PICK_MODE_CONTEXT *av2_alloc_pmc(const AV2_COMMON *cm, TREE_TYPE tree_type,
   return ctx;
 }
 
+// Resets the state of a PICK_MODE_CONTEXT node so it can be reused across
+// blocks/superblocks without re-allocation.
+void av2_reset_pmc(PICK_MODE_CONTEXT *ctx) {
+  if (ctx == NULL) return;
+  ctx->rd_mode_is_ready = 0;
+  av2_invalid_rd_stats(&ctx->rd_stats);
+  if (ctx->tx_type_map) av2_zero_array(ctx->tx_type_map, ctx->num_4x4_blk);
+  if (ctx->cctx_type_map)
+    av2_zero_array(ctx->cctx_type_map, ctx->num_4x4_blk_chroma);
+  ctx->skippable = 0;
+  ctx->hybrid_pred_diff = 0;
+  ctx->comp_pred_diff = 0;
+  ctx->single_pred_diff = 0;
+}
+
 void av2_free_pmc(PICK_MODE_CONTEXT *ctx, int num_planes) {
   if (ctx == NULL) return;
 
--- a/av2/encoder/encodeframe.c
+++ b/av2/encoder/encodeframe.c
@@ -817,9 +817,19 @@ static AVM_INLINE void encode_rd_sb(AV2_COMP *cpi, ThreadData *td,
       const BLOCK_SIZE min_partition_size = x->sb_enc.min_partition_size;
       init_encode_rd_sb(cpi, td, tile_data, sms_root, &dummy_rdc, mi_row,
                         mi_col, 1);
-      PC_TREE *const pc_root = av2_alloc_pc_tree_node(
-          xd->tree_type, mi_row, mi_col, cm->sb_size, sb_size, NULL,
-          PARTITION_NONE, 0, 1, ss_x, ss_y);
+      PC_TREE *pc_root;
+      if (cpi->sf.rt_sf.use_nonrd_partition) {
+        if (!td->pc_root) {
+          td->pc_root = av2_alloc_pc_tree_node(
+              xd->tree_type, mi_row, mi_col, cm->sb_size, sb_size, NULL,
+              PARTITION_NONE, 0, 1, ss_x, ss_y);
+        }
+        pc_root = td->pc_root;
+      } else {
+        pc_root = av2_alloc_pc_tree_node(xd->tree_type, mi_row, mi_col,
+                                         cm->sb_size, sb_size, NULL,
+                                         PARTITION_NONE, 0, 1, ss_x, ss_y);
+      }
       av2_reset_ptree_in_sbi(xd->sbi, xd->tree_type);
       av2_build_partition_tree_fixed_partitioning(
           cm, xd->tree_type, mi_row, mi_col,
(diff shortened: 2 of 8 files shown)


av2/encoder/context_tree.c
```diff
@@ -200,6 +200,21 @@ PICK_MODE_CONTEXT *av2_alloc_pmc(const AV2_COMMON *cm, TREE_TYPE tree_type,
...
 
+// Resets the state of a PICK_MODE_CONTEXT node so it can be reused across
+// blocks/superblocks without re-allocation.
+void av2_reset_pmc(PICK_MODE_CONTEXT *ctx) {
+  if (ctx == NULL) return;
+  ctx->rd_mode_is_ready = 0;
+  av2_invalid_rd_stats(&ctx->rd_stats);
+  if (ctx->tx_type_map) av2_zero_array(ctx->tx_type_map, ctx->num_4x4_blk);
+  if (ctx->cctx_type_map)
+    av2_zero_array(ctx->cctx_type_map, ctx->num_4x4_blk_chroma);
```
yeqing-wu, line 211: in this function, we only reset part of arrays, we did not reset per-plane arrays, would that be a problem?

```
  uint8_t *color_index_map[2];
  uint8_t *blk_skip[MAX_MB_PLANE];

  tran_low_t *coeff[MAX_MB_PLANE];
  tran_low_t *qcoeff[MAX_MB_PLANE];
  tran_low_t *dqcoeff[MAX_MB_PLANE];
  uint16_t *eobs[MAX_MB_PLANE];
  uint16_t *bobs[MAX_MB_PLANE];
  uint8_t *txb_entropy_ctx[MAX_MB_PLANE];
```

av2/encoder/ethread.c
```diff
@@ -427,6 +427,14 @@ static int enc_row_mt_worker_hook(void *arg1, void *unused) {
 
   assert(cur_tile_id != -1);
 
+  if (cpi->sf.rt_sf.use_nonrd_partition) {
+    thread_data->td->pc_root = av2_alloc_pc_tree_node(
+        SHARED_PART, 0, 0, cm->sb_size, cm->sb_size, NULL, PARTITION_NONE, 0, 1,
+        cm->seq_params.subsampling_x, cm->seq_params.subsampling_y);
```
yeqing-wu, line 436: this new allocation code would introduce the leak on setjmp/longjmp error path

av2/encoder/ethread.c
```diff
@@ -518,6 +529,14 @@ static int enc_worker_hook(void *arg1, void *unused) {
   }
   error_info->setjmp = 1;
 
+  if (cpi->sf.rt_sf.use_nonrd_partition) {
+    thread_data->td->pc_root = av2_alloc_pc_tree_node(
```
yeqing-wu, line 537: this new allocation code would introduce the leak on setjmp/longjmp error path

av2/encoder/encodeframe.c
```diff
@@ -2184,10 +2194,20 @@ static AVM_INLINE void encode_frame_internal(AV2_COMP *cpi) {
...
+    if (AVMMIN(mt_info->num_workers, cm->tiles.cols * cm->tiles.rows) > 1) {
       av2_encode_tiles_mt(cpi);
-    else
+    } else {
+      if (cpi->sf.rt_sf.use_nonrd_partition) {
+        cpi->td.pc_root = av2_alloc_pc_tree_node(
+            SHARED_PART, 0, 0, cm->sb_size, cm->sb_size, NULL, PARTITION_NONE,
+            0, 1, cm->seq_params.subsampling_x, cm->seq_params.subsampling_y);
+      } else {
+        cpi->td.pc_root = NULL;
```
yeqing-wu: overwrites pc_root = NULL without free, I think this would cause the leaks, right?

jianj-g: This is only allocated when `use_nonrd_partition` is active, and always freed below on line 2208.. so in this else branch `pc_root` is not overwriting existing memory.

av2/encoder/context_tree.c
jianj-g: This is not a problem for the following reasons:
    1. **`coeff` / `qcoeff` / `dqcoeff`**: These are not distinct buffers per node; they are pointer aliases to the thread's shared buffers (`shared_bufs->coeff_buf[i]`), and are overwritten dynamically during     
  transform search up to `eob`. In the initial `av2_alloc_pmc()`, they are not zero-initialized either.
    2. **`eobs` / `bobs` / `txb_entropy_ctx` / `color_index_map`**: In `av2_alloc_pmc()`, these are allocated via `avm_memalign()` without zeroing (`calloc`/`memset`). They are always explicitly written when a     
  mode/transform/palette is evaluated before being read.
    3. **`blk_skip`**: Fully written during mode search and fully overwritten in `av2_copy_tree_context()`.
    4. **State Gating**: `ctx->rd_mode_is_ready = 0` and `av2_invalid_rd_stats(&ctx->rd_stats)` ensure that a node's data is never read before a new mode is evaluated and selected.
    5. **Performance**: Zeroing all per-plane arrays across all superblocks would incur unnecessary `memset` memory bandwidth overhead and diminish the benefit of reusing the tree.

av2/encoder/ethread.c
jianj-g: Good catch! Thanks for the comment

av2/encoder/encodeframe.c
```diff
@@ -2184,10 +2194,20 @@ static AVM_INLINE void encode_frame_internal(AV2_COMP *cpi) {
...
+      }
       encode_tiles(cpi);
+      av2_free_pc_tree_recursive(cpi->td.pc_root, av2_num_planes(cm), 0, 0);
+      cpi->td.pc_root = NULL;
```
yeqing-wu, line 2207: Okay, if that’s the case, should we move the next line `cpi->td.pc_root = NULL;` into the `av2_free_pc_tree_recursive()` function to make the code cleaner? If `td.pc_root` is `NULL`, then the function can return quickly, and we don’t need to call `cpi->td.pc_root = NULL;` again. What are your thoughts on this?

jianj-g: Thanks for the suggestion. The function already returns early when it's NULL. Setting it inside the function would require passing it as a double pointer... And the other av2_free*() functions in the codebase don't do that so I think it's better to keep it this way.

Instead, I removed the redundant `else` branch above to avoid confusion.

yeqing-wu: okay to me since this is not big issue.

yeqing-wu (approved):

marpan-max (approved):

PR #5286, opened 2026-08-18
CCSO BO-first warm start under early termination

chengchen-google:
Evaluating BO-only first quickly establishes a competitive rate-distortion cost which might triggers more early termination.

For speed 2, RA, it provides coding gains (especially for chroma) with slightly speed up.

Speed2            PSNR-Y  PSNR-U  PSNR-V  PSNR-YUV   EncTime
A1 (17 frames)    -0.09%  -0.12%  -0.96%  -0.14%     99.36%
A2 (33 frames)    +0.02%  -1.43%  -1.31%  -0.10%     98.39%

STATS_CHANGED

Change-Id: Ied7c73b086f8f475dd8db6a07534f2449a0a5452


yeqing-wu (approved):
nice speedup with coding gain

PR #5289, opened 2026-08-19
Reduce intra mode top count to 3 for speed 3 and 4

chengchen-google:
By reducing the number of candidates, we achieve a considerable
speed up with accepted coding loss.

Result of reducing the count to 3 for speed 4:

TestSet          PSNR-YUV   EncTime    Ratio
A1 (17 frames)    0.10%      96.70%    34.1
A2 (33 frames)    0.07%      96.81%    47.0

Result of reducing the count to 3 for speed 3:

TestSet          PSNR-YUV   EncTime    Ratio
A1 (17 frames)    0.11%      96.87%    29.4
A2 (33 frames)    0.09%      96.41%    41.3

If we reduce the count to 2
The speedup/loss ratio is good for A2, but not good enough for A1.
On speed 4, RA, the performance is:

TestSet          PSNR-YUV   EncTime    Ratio
A1 (17 frames)    0.41%      92.76%    19.0
A2 (33 frames)    0.19%      93.44%    36.9

STATS_CHANGED

### DIFF
--- a/av2/common/pred_common.c
+++ b/av2/common/pred_common.c
@@ -798,12 +798,13 @@ int av2_get_cdef_context(const AV2_COMMON *const cm,
   const int block_mask = ~(cdef_size - 1);
 
   const CommonModeInfoParams *const mi_params = &cm->mi_params;
+  const TileInfo *const tile = &xd->tile;
   const int mi_row = xd->mi_row;
   const int mi_col = xd->mi_col;
 
   const int left_cdef_mi_col = mi_col - cdef_size;
   MB_MODE_INFO *neighbor0 = NULL;
-  if (left_cdef_mi_col >= 0) {
+  if (left_cdef_mi_col >= tile->mi_col_start) {
     const int left_grid_idx = get_mi_grid_idx(mi_params, mi_row & block_mask,
                                               left_cdef_mi_col & block_mask);
     neighbor0 = mi_params->mi_grid_base[left_grid_idx];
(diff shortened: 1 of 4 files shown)


urvangjoshi:
The ratio for A1 set seems less than the threshold, right? 
What if you used value 3 instead of 2?

chengchen-google:
> The ratio for A1 set seems less than the threshold, right? What if you used value 3 instead of 2?

Yes, you are right. I am testing 3.

leolzhao:
The trade off for class A1 is a little bit worse than the threshold, what about enabling this speed feature for speed 5 and above? @chengchen-google

leolzhao:
> > The ratio for A1 set seems less than the threshold, right? What if you used value 3 instead of 2?
> 
> Yes, you are right. I am testing 3.

if 3 is good for cpu-used 4, we can set it to 3 for speed 4 and 2 for speed 5 and above.

chengchen-google:
I tested to reduce the count to 3 and tested on speed 3 and 4. The tradeoff is better. Please see the updated commit message.

yeqing-wu:
What is the anchor of the results? Meanwhile, should we remove 'Change-Id: Ib9dba881748b9106f295f4d468cdbf0367e28a7d'?

chengchen-google:
> What is the anchor of the results? Meanwhile, should we remove 'Change-Id: Ib9dba881748b9106f295f4d468cdbf0367e28a7d'?

The anchor is ea89c216c629019f586ced5a701e56200169e012.

leolzhao (approved):

Issue #5234, opened 2026-08-10
CONFIG_12BIT_PROFILE: New profile missing from multistream conformance checks


peterderivaz:
CONFIG_12BIT_PROFILE add support for new profiles.

However, there is still some code that has not been updated, e.g. in conformance_check_msdo_lcr:

```
  if (num_extended_layers > 1 && num_embedded_layers == 1) {
    if (msdo_present &&
        (msdo_prof == MAIN_420_10_IP0 || msdo_prof == MAIN_420_10_IP1 ||
         msdo_prof == MAIN_420_10_IP2 || msdo_prof == MAIN_422_10_IP1 ||
         msdo_prof == MAIN_444_10_IP1))
      return true;

    if (global_lcr_present && glcr_prof == MAIN_420_10_IP2) return true;
  }

  if (num_extended_layers == 1 && num_embedded_layers > 1) {
    if (!msdo_present && local_lcr_present &&
        (llcr_prof == MAIN_420_10_IP1 || llcr_prof == MAIN_422_10_IP1 ||
         llcr_prof == MAIN_444_10_IP1))
      return true;

    if (!msdo_present && (global_lcr_present || local_lcr_present) &&
        (glcr_prof == MAIN_420_10_IP2 || llcr_prof == MAIN_420_10_IP2))
      return true;
  }
  ...
  return false;
```

The problem is that if we use the new profile, then it is impossible to make a legal multistream.

This is because only certain explicit profiles are detected as being legal.

**Suggested Fix**

Use a helper function (does one already exist?) that extracts the IP number, and make these checks depend on IP number instead of explicit profiles.

xinzhao-apple:
Thanks @peterderivaz, I will submit a fix to this.

PR #5290, opened 2026-08-19
Fix MSDO/LCR conformance check for the 12-bit profile

xinzhao-apple:
conformance_check_msdo_lcr() compared against enumerated profile labels, which were never updated for MAIN_444C_12_IP2, so no legal multistream was possible with the 12-bit profile. Add av2_get_interop_from_profile() and express the checks in terms of interoperability points instead.

Also fixes two related defects: get_lcr_global_interop() returned a profile_idc where lcr_max_interop is an IOP, and the single-xlayer/multi-mlayer check OR'd the LCR presence flags, letting the absent-LCR fallback admit an IOP 0 stream.

Fix issue #5234.

### DIFF
--- a/av2/common/annexA.c
+++ b/av2/common/annexA.c
@@ -84,14 +84,6 @@ typedef enum {
 */
 /* clang-format on */
 
-typedef enum {
-  INTEROP_0,
-  INTEROP_1,
-  INTEROP_2,
-  INTEROP_3,
-  NUM_INTEROP_POINTS = 16,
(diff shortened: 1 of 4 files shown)


SMPaluri (approved):

peterderivaz (approved):
James has tested the failing test streams and they pass now, thanks!

xinzhao-apple:
> James has tested the failing test streams and they pass now, thanks!

Great, thanks for the update!

PR #5296, opened 2026-08-20
Refactor compound type search functions

RemyaPrakasan:


RemyaPrakasan:
@yunqingwang1 @urvangjoshi
/cc @vikasprasad10 @harishdm @ranjit-tulabandu @cherma-a @deepa-kg @ram-mohan

We have verified that this PR is bit-exact for speeds 0 to 3 across RA and LD configurations on a few test clips. Please review.

Note: Each commit in this PR has its own commit message. Please do not squash them.

yunqingwang1 (approved):
Looks good - thanks.

Issue #5297, opened 2026-08-20
Encoder segfault with --min-gf-interval below MIN_GF_INTERVAL


xinzhao-apple:
### What version / commit were you testing with?

`git describe`  or `git log -1` can produce this info.

v1.0.0-85-g8a2a0a92c

### What steps will reproduce the problem?

1. Build `avmenc` in a release configuration.
2. Run "avmenc --min-gf-interval=1 --max-gf-interval=2 --end-usage=cbr --cpu-used=4 --limit=2 -o out.ivf Vertical_Bayshore_270x480_2997.y4m"
3. The encoder segfaults while encoding the **last** frame of the sequence. Reproduces deterministically. `--min-gf-interval=1`, `2` and `3` all crash;`--min-gf-interval=4` (or `0`, meaning auto) is fine. `--max-gf-interval` is not involved, `--max-gf-interval=0 --min-gf-interval=1` also crashes.

### Expected behavior

Either the encode completes, or `avmenc` rejects the out-of-range value with a diagnostic. The CLI accepts `--min-gf-interval=1` without complaint.

### Actual behavior

Error message:
POC:     0 [ KEY ][Level:1][Q: 40]:      42365 Bytes, 81650.2ms    [ -1, -1, -1, -1, -1, -1, -1,] 270x480
zsh: segmentation fault  ./avmenc --min-gf-interval=1 --max-gf-interval=2 --end-usage=cbr --cpu-used=4

PR #5298, opened 2026-08-20
Add range check min_gf_interval in validate_config

xinzhao-apple:
--min-gf-interval=1..3 bypasses the clamp the auto path applies, which defeats the baseline_gf_interval < min_gf_interval guard in define_gf_group(). A degenerate end-of-sequence group then builds an ARF whose temporal filter reads past the end of the lookahead and segfaults.

Valid configurations are bit-identical; 1..3 now match min=4.

This resolves issue #5297.


urvangjoshi:
@xinzhao-apple :  Good catch!  

However, I see that for max_gf_interval, we have a range check here itself: https://github.com/AOMediaCodec/avm/blob/main/av2/av2_cx_iface.c#L698

It seems best to do the same for min_gf_interval also.

xinzhao-apple:
> @xinzhao-apple : Good catch!
> 
> However, I see that for max_gf_interval, we have a range check here itself: https://github.com/AOMediaCodec/avm/blob/main/av2/av2_cx_iface.c#L698
> 
> It seems best to do the same for min_gf_interval also.

@urvangjoshi , I agree, the commit has been amended, please take a look. Thanks!

urvangjoshi (approved):

PR #5304, opened 2026-08-21
Bound the two-pass wet-pass mode search with the dry pass's rd

yeqing-wu:
The wet pass starts mode search from an unbounded rd limit, so its pruning cannot bite until it finds a good candidate on its own. Start it instead from the dry pass's rd for the same block and size, plus a margin.

The two passes see different contexts, so the bound can be violated. A block left with no valid mode is re-searched once with the original limit.

Anchor: commit c77f169

```
  | A1         | -0.30 | -0.29 | -0.38 | -0.30 |  100 |  100 |
  | A2         | -0.17 |  0.07 |  0.31 | -0.15 |  101 |  100 |
  | Avg w/o B2 | -0.21 | -0.04 |  0.11 | -0.19 |  100 |  100 |
```

STATS_CHANGED

### DIFF
--- a/av2/encoder/block.h
+++ b/av2/encoder/block.h
@@ -30,6 +30,19 @@
 extern "C" {
 #endif
 
+/*! \brief Granularity of the dry-pass rd grid, in mi units (log2): 16x16 px. */
+#define TWO_PASS_DRY_RD_UNIT_LOG2 2
+/*! \brief Side of the dry-pass rd grid, in units. */
+#define TWO_PASS_DRY_RD_SIDE (MAX_MIB_SIZE >> TWO_PASS_DRY_RD_UNIT_LOG2)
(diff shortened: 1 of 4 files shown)


yeqing-wu:
@urvangjoshi,   Update on reusing the MV from the dry pass as the starting MV of the wet pass: the gain has dropped from -0.13% to -0.03%. Part of the reason is that the anchor itself increased an additional -2% gain compared to before.

Meanwhile, there have also been a lot of recent commits (several for MV search and DRL), so I'm still going through them to check whether anything was missed, and to see if I can push the gain further. Just let you know.

yeqing-wu:
Hello, can anyone help review the code?

jingninghan (approved):

leolzhao (approved):

PR #5308, opened 2026-08-24
Temporal filter function optimization

yunqingwang1:
Pulled common calculations from the pixel filtering loop and conducted them beforehand.

CTC tests showed no stats change.


yeqing-wu (approved):
LGTM

PR #5313, opened 2026-08-25
Use encoder-only method for reduced-tx-part-set option

urvangjoshi:
- Always enable all tx types at sequence level
- Disable 4way/5way tx types only during RDO

Fixes https://github.com/AOMediaCodec/av2-spec-internal/issues/1262 using encoder-only method.

### DIFF
--- a/av2/encoder/encoder.c
+++ b/av2/encoder/encoder.c
@@ -577,6 +577,11 @@ void av2_init_seq_coding_tools(AV2_COMP *cpi, SequenceHeader *seq,
   seq->enable_mhccp = oxcf->intra_mode_cfg.enable_mhccp;
   seq->enable_inter_ddt =
       seq->single_picture_header_flag ? 0 : oxcf->txfm_cfg.enable_inter_ddt;
+  // Note: we always keep the sequence header flag `reduced_tx_part_set` off,
+  // ignoring the encoder option `oxcf->txfm_cfg.reduced_tx_part_set`. This is
+  // because the restricting the transform partition types during encoder RDO
+  // only has been found to be more compression-efficient than using the
+  // sequence header option.
   seq->reduced_tx_part_set = 0;
   seq->enable_cctx = oxcf->txfm_cfg.enable_cctx;
   seq->enable_ibp = oxcf->intra_mode_cfg.enable_ibp;
--- a/av2/encoder/encoder.c
+++ b/av2/encoder/encoder.c
@@ -577,7 +577,7 @@ void av2_init_seq_coding_tools(AV2_COMP *cpi, SequenceHeader *seq,
   seq->enable_mhccp = oxcf->intra_mode_cfg.enable_mhccp;
   seq->enable_inter_ddt =
       seq->single_picture_header_flag ? 0 : oxcf->txfm_cfg.enable_inter_ddt;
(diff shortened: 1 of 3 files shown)


av2/encoder/encoder.c
```diff
@@ -577,7 +577,7 @@ void av2_init_seq_coding_tools(AV2_COMP *cpi, SequenceHeader *seq,
   seq->enable_mhccp = oxcf->intra_mode_cfg.enable_mhccp;
   seq->enable_inter_ddt =
       seq->single_picture_header_flag ? 0 : oxcf->txfm_cfg.enable_inter_ddt;
-  seq->reduced_tx_part_set = oxcf->txfm_cfg.reduced_tx_part_set;
+  seq->reduced_tx_part_set = 0;
```
wantehchang, line 585: Please add a comment to explain why we ignore `oxcf->txfm_cfg.reduced_tx_part_set` and set `seq->reduced_tx_part_set` to 0.

Another good place to add a comment is in av2/encoder/encoder.h, where we can document where `oxcf->txfm_cfg.reduced_tx_part_set` should be used:

```
  /*!
   * Flag to indicate if reduced transform block partition set should be
   * enabled.
   */
  bool reduced_tx_part_set;
```

av2/encoder/tx_search.c
```diff
@@ -3048,8 +3048,7 @@ static void select_tx_partition_type(
   uint8_t full_blk_skip[MAX_TX_PARTITIONS] = { 0 };
 
   for (TX_PARTITION_TYPE type = 0; type < TX_PARTITION_TYPES; ++type) {
-    if (cpi->common.seq_params.reduced_tx_part_set &&
-        type > TX_PARTITION_VERT) {
+    if (cpi->oxcf.txfm_cfg.reduced_tx_part_set && type > TX_PARTITION_VERT) {
```
wantehchang, line 3051: I wanted to confirm that the following line in av2/encoder/partition_search.c should continue to use `cm->seq_params.reduced_tx_part_set`, not `cpi->oxcf.txfm_cfg.reduced_tx_part_set`:

```
static void pick_sb_modes(AV2_COMP *const cpi, ThreadData *td,
                          TileDataEnc *tile_data, MACROBLOCK *const x,
                          int mi_row, int mi_col, RD_STATS *rd_cost,
                          PARTITION_TYPE partition, REGION_TYPE cur_region_type,
                          int sb_root_partition_info, BLOCK_SIZE bsize,
                          PICK_MODE_CONTEXT *ctx, RD_STATS best_rd) {
  ...
  xd->reduced_tx_part_set = cm->seq_params.reduced_tx_part_set;
```

The reason is not obvious. A comment at some appropriate place would be helpful.

av2/encoder/encoder.c
urvangjoshi: Thanks!  Added a comment.

av2/encoder/tx_search.c
urvangjoshi: Yes, `xd->reduced_tx_part_set` is purely used for signaling  (and signaling cost computation) in both encoder and decoder.  So, that initialization is correct.

I added a comment where `reduced_tx_part_set` is declared in `xd`.

urvangjoshi (commented):
Thanks for the review! PTAL.

wantehchang (approved):

leolzhao (approved):
LGTM.

PR #5322, opened 2026-08-25
Check obu_type if single_picture_header_flag is 1

### LINKED ISSUE #5319: In a still-picture sequence, the decoder crashes when decoding a second frame of type OBU_BRIDGE_FRAME
cc: @peterderivaz @urvangjoshi 

### What version / commit were you testing with?

v1.0.0-90-g4bdc67918 (commit 4bdc679181083f5ae0102a48ad07392b93b85fc6)

### What steps will reproduce the problem?

This bug was detected by a libavif fuzz test and was originally reported in Google internal bug b/552470033.

The stack trace of the crash is:

```
CRASH LOG:     @     0x7fa68fe39350  (unknown)  (unknown)
CRASH LOG:     @     0x55b0ca1abc06       5600  avm/av2/decoder/decodeframe.c:8451 read_uncompressed_header
CRASH LOG:     @     0x55b0ca1a4df4        320  avm/av2/decoder/decodeframe.c:9572 av2_read_tilegroup_header
CRASH LOG:     @     0x55b0ca278203       1520  avm/av2/decoder/obu.c:773 avm_decode_frame_from_obus
CRASH LOG:     @     0x55b0ca264695        976  avm/av2/decoder/decoder.c:859 av2_receive_compressed_data
CRASH LOG:     @     0x55b0ca0e3546         48  avm/av2/av2_dx_iface.c:398 frame_worker_hook
CRASH LOG:     @     0x55b0ca653418         48  execute
CRASH LOG:     @     0x55b0ca0dcfb6       3632  avm/av2/av2_dx_iface.c:546 decoder_decode
CRASH LOG:     @     0x55b0ca53e1d8         80  avm_codec_decode
CRASH LOG:     @     0x55b0c970270c        304  avmCodecGetNextImage
CRASH LOG:     @     0x55b0c96ed6a0        368  libavif/src/read.c:6794 avifDecoderNextImage
CRASH LOG:     @     0x55b0c96ef46d         48  libavif/src/read.c:7305 avifDecoderRead
```

The debugger shows the following at the crash:

```
* thread #1, name = 'avif_fuzztest_d', stop reason = signal SIGILL: illegal operand
       frame #0: 0x00007ffebe93d74f setup_frame_size(cm=0x000011ae7
a4b3280, frame_size_override_flag=0, rb=0x00007fffffff4ae8) at decodeframe.c:3990:71
   3987	    const RefCntBuffer *ref_buf = get_ref_frame_buf(
   3988	        cm, cm->bridge_frame_info.bridge_frame_ref_idx_remapped);
   3989	    width =
-> 3990	        AVMMIN(cm->bridge_frame_info.bridge_frame_max_width, ref_buf->width);
   3991	    height =
   3992	        AVMMIN(cm->bridge_frame_info.bridge_frame_max_height, ref_buf->height);
   3993	  } else {
(lldb) p cm->bridge_frame_info.bridge_frame_ref_idx_remapped
(int) 0
(lldb) p ref_buf
(const RefCntBuffer *) NULL
(lldb)
```

In the debugger I see the following:

The first OBU has an invalid OBU type and is ignored.

The second OBU is a sequence header OBU with single_picture_header_flag=1.

The third OBU is an OBU_OPEN_LOOP_KEY.

The fourth OBU is an OBU_BRIDGE_FRAME. It is this OBU that causes the crash because single_picture_header_flag=1 causes its FrameType to be set to KEY_FRAME, and things go wrong from there.

Since single_picture_header_flag=1 implies still_picture=1, the coded video sequence contains only one coded frame. So we should reject the fourth OBU on that ground. But I haven't figured out how to count the number of coded frames in a coded video sequence. So I will write a pull request to reject the fourth OBU by its OBU type -- the OBU type of a KEY_FRAME must be either OBU_CLOSED_LOOP_KEY or OBU_OPEN_LOOP_KEY.

wantehchang:
If single_picture_header_flag is equal to 1 in the sequence header OBU, frame_type is inferred to be KEY_FRAME and immediate_output_frame is inferred to be 1. Therefore obu_type needs to be OBU_CLOSED_LOOP_KEY, otherwise frame_type and immediate_output_frame will be inconsistent with obu_type and things will go wrong.

Fixes https://github.com/AOMediaCodec/avm/issues/5319.

### DIFF
--- a/av2/decoder/decodeframe.c
+++ b/av2/decoder/decodeframe.c
@@ -8007,6 +8007,11 @@ static int read_uncompressed_header(AV2Decoder *pbi, OBU_TYPE obu_type,
     cm->implicit_output_picture = 0;
     cm->cur_frame->immediate_output_picture = 1;
     cm->cur_frame->implicit_output_picture = 0;
+    if (obu_type != OBU_CLOSED_LOOP_KEY) {
+      avm_internal_error(
+          &cm->error, AVM_CODEC_CORRUPT_FRAME,
+          "Still pictures must be coded as closed loop keyframes");
+    }
     current_frame->frame_type = KEY_FRAME;
     pbi->decoding_first_frame = 1;
     reset_frame_buffers(cm);


av2/decoder/decodeframe.c
wantehchang, line 8026: In the `single_picture_header_flag=0` case, `OBU_CLOSED_LOOP_KEY` and `OBU_OPEN_LOOP_KEY` are mapped to `KEY_FRAME`.

wantehchang, line 8112: In the `single_picture_header_flag=0` case, if `still_picture` is equal to 1, we require the OBU type be `OBU_CLOSED_LOOP_KEY` or `OBU_OPEN_LOOP_KEY` via the `current_frame->frame_type != KEY_FRAME` check. The check I added at line 8010 for the `single_picture_header_flag=1` case matches this check.

However, I am not sure if we should allow `OBU_OPEN_LOOP_KEY` in a still-picture sequence. What do you think?

av2/decoder/decodeframe.c
```diff
@@ -8007,6 +8007,10 @@ static int read_uncompressed_header(AV2Decoder *pbi, OBU_TYPE obu_type,
     cm->implicit_output_picture = 0;
     cm->cur_frame->immediate_output_picture = 1;
     cm->cur_frame->implicit_output_picture = 0;
+    if (obu_type != OBU_CLOSED_LOOP_KEY && obu_type != OBU_OPEN_LOOP_KEY) {
+      avm_internal_error(&cm->error, AVM_CODEC_CORRUPT_FRAME,
+                         "Still pictures must be coded as keyframes");
+    }
```
wantehchang, line 8014: Peter, Urvang: Do you know of a way to check if the still-picture sequence has more than one frame?

urvangjoshi: @peterderivaz ?

av2/decoder/decodeframe.c
peterderivaz: With single picture header flag we also infer:

```
if ( single_picture_header_flag ) {
        ShowExistingFrame = 0
        FrameType = KEY_FRAME
        FrameIsIntra = 1
        immediate_output_frame = 1
        implicit_output_frame = 0
```

Compare with the normal inference of immediate_output_frame:

```
       if ( IsBridge || obu_type == OBU_OPEN_LOOP_KEY ) {
            immediate_output_frame = 0
        } else {
            f(1) immediate_output_frame;
        }
```

So it seems like OBU_OPEN_LOOP_KEY should not be allowed with single_picture_header_flag as it would result in an inconsistent value of immediate_output_frame.

av2/decoder/decodeframe.c
peterderivaz: Not sure in the reference code.  In the spec it says:

> **still_picture** equal to 1 specifies that the coded video sequence contains only one [=coded frame=].

but I am not aware of any place in the reference code where this is checked.

peterderivaz:
I was wondering whether this is also a bug with the specification and if we should add a corresponding bitstream conformance check into the specification to match the code changes.

However, on reflection, the specification does say that:
1. single_picture_header_flag equal to 1 means still_picture is equal to 1
2. still_picture equal to 1 means the CVS has one frame
3. CVS has one frame means that it starts with OBU_CLOSED_LOOP_KEY (due to the following text)

> A new coded video sequence for an extended layer is defined to start at each temporal unit that contains an OBU with obu_type equal to OBU_CLOSED_LOOP_KEY in the coded extended layer unit corresponding to the extended layer.

So I think that the specification is okay.

av2/decoder/decodeframe.c
wantehchang: Peter, Urvang: I just realized that AV2 allows multiple still-picture frames to share the same sequence header OBU, so the decoder needs to allow this kind of bitstream:

```
[OBU_SEQUENCE_HEADER seq_header_id=0 still_picture=1] 
[OBU_CLOSED_LOOP_KEY cur_mfh_id=0 seq_header_id_in_frame_header=0] 
[OBU_CLOSED_LOOP_KEY cur_mfh_id=0 seq_header_id_in_frame_header=0]
```

Instead of checking whether a still-picture sequence has more than one frame, I think a decoder should consider a new coded video sequence starts whenever a frame activates a sequence header OBU with still_picture=1.

What do you think?

av2/decoder/decodeframe.c
wantehchang: Peter: I agree with your analysis. The following requirement is also relevant:

> If still_picture is equal to 1, it is a requirement of bitstream conformance that FrameType is equal to KEY_FRAME and immediate_output_frame is equal to 1.

The requirement you cited in https://github.com/AOMediaCodec/avm/pull/5322#issuecomment-5442897393 also means only `OBU_CLOSED_LOOP_KEY` is allowed.

I will disallow `OBU_OPEN_LOOP_KEY`.

urvangjoshi (approved):

PR #5323, opened 2026-08-25
Pre-allocate CCSO search buffers

YixinDu1:
Replace per-frame calloc/free in av2_ccso_search with persistent buffers on AV2_COMP, freed at encoder close. No bitstream impact.


tianqitliu:
This commit looks good to me.
It converts all per-frame CCSO search buffers into persistent AV2_COMP buffers, so the encoder no longer needs to reallocate them every frame. It also fixes the over-allocation of the chroma buffers, which were previously sized at full luma resolution.

urvangjoshi (approved):

PR #5324, opened 2026-08-25
Add SMS-based partition pre-screener for inter frames

YixinDu1:
Add a unified SMS pre-screener (sms_unified_prune) that runs one MLP inference per square block before the partition search and uses the output to prune partition candidates.

Anchor: da53762
Test condition: CTC, RA, 33 frames, speed 4

RA:
```
+----------------+------+------+------+------+------+
| Class          |    Y |   Cb |   Cr | wAvg | Enc% |
+----------------+------+------+------+------+------+
| A1             | 0.01 | 0.35 |-0.09 | 0.02 | 96.3 |
| A2             | 0.14 | 0.14 | 0.27 | 0.14 | 95.5 |
| Average w/o B2 | 0.09 | 0.12 | 0.29 | 0.10 | 96.7 |
+----------------+------+------+------+------+------+
```
STATS_CHANGED


av2/encoder/partition_sms.c
```diff
@@ -0,0 +1,315 @@
...
+  (void)mi_row;
+  (void)mi_col;
+}
+
+/* -------------------------------------------------------------------
+ * MLP forward pass (2 hidden layers, ReLU)
+ * Weights absorb LayerNorm and z-score normalization (see export script).
+ * ------------------------------------------------------------------- */
+
+static void mlp_forward(const float *in, int in_dim, const float *w1,
```
urvangjoshi: We already have `struct NN_CONFIG`  and relevant functions like `av2_nn_predict` to handle DNN / MLP networks.
There is also integration with TFLite -- which is used by certain functions.
So, let's not add new APIs for this -- pls reuse one of the existing APIs.

av2/encoder/partition_sms.h
```diff
@@ -0,0 +1,40 @@
...
+
+/* Run MLP inference for the block at (mi_row, mi_col) with the given bsize.
+ * Stores per-class softmax probabilities in sms_tree->sms_unified_probs[],
+ * indexed by PARTITION_TYPE enum value. Must be called before any pruning
+ * gates that consume sms_unified_probs. */
+void av2_sms_unified_compute(AV2_COMP *const cpi, MACROBLOCK *x,
+                             SIMPLE_MOTION_DATA_TREE *sms_tree, int mi_row,
+                             int mi_col, BLOCK_SIZE bsize);
+
+/* Prune HORZ partition using the probabilities computed by
```
urvangjoshi, line 29: Any reason to prune HORZ only? And not VERT?

urvangjoshi (commented):
@YixinDu1 : Thanks for the PR.  The loss is slightly higher for A2, but there's speed-up as well.
Question: what are the numbers for lower resolution sets?  I see that you disabled the speed feature for some lower resolutions.

YixinDu1:
> @YixinDu1 : Thanks for the PR. The loss is slightly higher for A2, but there's speed-up as well. Question: what are the numbers for lower resolution sets? I see that you disabled the speed feature for some lower resolutions.

Thanks for the comment. The loss for lower res is around 0.1x and there is around 2% speedup.

av2/encoder/partition_sms.h
YixinDu1: The current PR only prunes HORZ — we trained a VERT model as well, but its recall on VERT partitions was insufficient to meet our quality bar, so it was left out for now. We plan to revisit in a follow-up once we have a stronger VERT model.

av2/encoder/context_tree.h
```diff
@@ -108,6 +108,11 @@ typedef struct SIMPLE_MOTION_DATA_TREE {
   unsigned int sms_rect_feat[8];
   int sms_none_valid;
   int sms_rect_valid;
+
+  // Unified SMS pre-screener: computed once per block and reused at each
+  // pruning gate.
+  float sms_unified_probs[10];
```
urvangjoshi: What is the 10 here?  Use a `#define` pls

av2/encoder/partition_sms.c
```diff
@@ -0,0 +1,281 @@
...
+    default: return -1;
+  }
+}
+
+/* -------------------------------------------------------------------
+ * Feature extraction
+ * ------------------------------------------------------------------- */
+
+#define SMS_FEAT_DIM 52
+
```
urvangjoshi, line 52: Add a comment here explaining what and how many features are extracted by this function.

av2/encoder/partition_sms.c
```diff
@@ -0,0 +1,281 @@
...
+                                 int mi_row, int mi_col, BLOCK_SIZE bsize,
+                                 float *feat) {
+  const AV2_COMMON *const cm = &cpi->common;
+  const MACROBLOCKD *const xd = &x->e_mbd;
+  int f = 0;
+
+  const int ref = get_closest_pastcur_ref_or_ref0(cm);
+
+  /* Gather sub-block pointers */
+  const SIMPLE_MOTION_DATA_TREE *q[4];
```
urvangjoshi: Do we have a `#define` for `4`?

av2/encoder/partition_sms.c
```diff
@@ -0,0 +1,281 @@
...
+  const BLOCK_SIZE left_bs =
+      has_left ? xd->left_mbmi->sb_type[xd->tree_type == CHROMA_PART] : bsize;
+  feat[f++] = (float)has_above;
+  feat[f++] = (float)mi_size_wide_log2[above_bs];
+  feat[f++] = (float)mi_size_high_log2[above_bs];
+  feat[f++] = (float)has_left;
+  feat[f++] = (float)mi_size_wide_log2[left_bs];
+  feat[f++] = (float)mi_size_high_log2[left_bs];
+
+  feat[f++] = (float)((AVMMIN(cm->width, cm->height) >= 480) +
```
urvangjoshi: Use a local const for `AVMMIN(cm->width, cm->height)` to avoid repetition.

av2/encoder/partition_sms.c
```diff
@@ -0,0 +1,281 @@
...
+  const int half_h = mi_size_high[bsize] / 2;
+  int left_mid_horz = 0;
+  if (xd->left_available && half_h > 0 &&
+      mi_row + half_h < cm->mi_params.mi_rows) {
+    const MB_MODE_INFO *m_top = xd->mi[(half_h - 1) * xd->mi_stride - 1];
+    const MB_MODE_INFO *m_bot = xd->mi[half_h * xd->mi_stride - 1];
+    if (m_top && m_bot)
+      left_mid_horz = (m_top->mi_row_start != m_bot->mi_row_start ||
+                       m_top->mi_col_start != m_bot->mi_col_start)
+                          ? 1
```
urvangjoshi: `? 1 : 0` part is redundant. Pls remove.

av2/encoder/partition_sms.c
```diff
@@ -0,0 +1,281 @@
...
+  avm_clear_system_state();
+
+  const int slot = sms_bsize_slot(bsize);
+  if (slot < 0) return;
+  const NN_CONFIG *nn_config = &sms_nn_configs[slot];
+
+  float feat[SMS_FEAT_DIM];
+  extract_sms_features(cpi, x, sms_tree, mi_row, mi_col, bsize, feat);
+
+  av2_nn_predict(feat, nn_config, 0, sms_tree->sms_unified_probs);
```
urvangjoshi, line 270: We always call this function with `reduce_prec = 1`.   This ensures consistency across platforms. Pls do the same here.

av2/encoder/partition_sms.c
```diff
@@ -0,0 +1,281 @@
...
+                                SIMPLE_MOTION_DATA_TREE *sms_tree,
+                                PartitionSearchState *part_search_state) {
+  if (!sms_tree || !sms_tree->sms_unified_valid) return;
+  if (cpi->is_screen_content_type) return;
+
+  const float *probs = sms_tree->sms_unified_probs;
+  const int slot = sms_bsize_slot(sms_tree->block_size);
+  if (slot < 0) return;
+
+  const int horz_qualifies =
```
urvangjoshi: Suggest using `bool` instead

av2/encoder/partition_sms.c
YixinDu1: Refactored to use the existing APIs

av2/encoder/context_tree.h
YixinDu1: Done

av2/encoder/partition_sms.c
YixinDu1: Added comments

av2/encoder/partition_sms.c
YixinDu1: Done

av2/encoder/partition_sms.c
YixinDu1: Fixed

av2/encoder/partition_sms.c
YixinDu1: Fixed

av2/encoder/partition_sms.c
YixinDu1: Fixed

av2/encoder/partition_sms.c
YixinDu1: Fixed

av2/encoder/partition_sms.c
YixinDu1: With reduce_prec=1, the UBSan CI unit test fails: av2_nn_output_prec_reduce computes (int)(output[i] * 512 + 0.5), but our models have extreme weight magnitudes that produce logits on the order of 35M — 35M × 512 ≈ 18B, which overflows int and triggers undefined behavior. We've reverted to reduce_prec=0 to fix this. Since the logits are immediately consumed by av2_nn_softmax (which is shift-invariant), any cross-platform difference in raw logit values cancels out completely after normalization, so reduce_prec=0 has no effect on output consistency. For future model development, we will ensure weights are scaled such that reduce_prec=1 can be used safely. Are you OK with this?

urvangjoshi: I see.   Yeah, should be fine as long as encoder output is consistent over platforms.

urvangjoshi (approved):

PR #5329, opened 2026-08-26
Optimize TCQ with block-level AVX2 dispatch

chengchen-google:
Previously, TCQ evaluated coefficients along anti-diagonal scan lines in a portable C loop (trellis_loop_diagonal_st8), invoking individual fine-grained RTCD SIMD functions per coefficient.

This fine-grained dispatch structure incurred significant overhead:
- Repeated RTCD function calls and indirect branches on every coeff.
- Register spills to memory structures across subroutine boundaries.
- Optimization barriers preventing compiler vector scheduling across the full diagonal loop.

This patch introduces a coarse-grained, block-level RTCD entry point:
  av2_trellis_loop_diagonal_st8()
with AVX2 specialization (av2_trellis_loop_diagonal_st8_avx2).

Updates are now executed contiguously within AVX2 registers without exiting to scalar C code per coefficient.

Unit tests are added to compare the the new and original implementation.

We can see that the TCQ function becomes 3-10% faster.

```
=======================================================================
TX Size   Pure C (us)   Base AVX2 (us)   Patch AVX2 (us)  vs Base AVX2
=======================================================================
4x4       110313        45646            44360           1.03x (+2.8%)
8x8       369747        102636           96537           1.06x (+5.9%)
16x16     343618        81354            74201           1.10x (+8.8%)
32x32     268794        60375            54646           1.10x (+9.5%)
4x8       195707        64939            61781           1.05x (+4.9%)
8x4       201683        65227            61872           1.05x (+5.1%)
8x16      175484        44150            40669           1.09x (+7.9%)
16x8      180465        44455            40943           1.09x (+7.9%)
16x32     134924        30975            28011           1.11x (+9.6%)
32x16     134311        30946            28192           1.10x (+8.9%)
4x16      364581        103343           96494           1.07x (+6.6%)
16x4      390933        104326           97702           1.07x (+6.3%)
8x32      343005        81586            74379           1.10x (+8.8%)
32x8      352547        81916            74820           1.09x (+8.7%)
=======================================================================
```

### DIFF
--- a/test/trellis_test.cc
+++ b/test/trellis_test.cc
@@ -17,6 +17,7 @@
 #include "third_party/googletest/src/googletest/include/gtest/gtest.h"
 #include "test/register_state_check.h"
 #include "test/function_equivalence_test.h"
+#include "test/util.h"
 
 #include "config/avm_config.h"
 #include "config/avm_dsp_rtcd.h"
@@ -24,7 +25,10 @@
 
 #include "avm/avm_integer.h"
 #include "av2/common/enums.h"
+#include "av2/common/idct.h"
+#include "av2/common/scan.h"
 #include "av2/encoder/trellis_quant.h"
+#include "av2/common/txb_common.h"
 
 using libavm_test::FunctionEquivalenceTest;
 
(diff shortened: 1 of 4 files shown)


test/trellis_test.cc
```diff
@@ -878,6 +882,503 @@ class TcqDecideStatesQ1Test
...
+                                            const qm_val_t *iqmatrix) {
+    int dqv = dequant[coeff_idx != 0];
+    if (iqmatrix != NULL) {
+      dqv =
+          (dqv * iqmatrix[coeff_idx] + (1 << (AVM_QM_BITS - 1))) >> AVM_QM_BITS;
+    }
+    return dqv;
+  }
+
+  static void trellis_loop_diagonal_st8_prepatch_baseline_avx2(
```
jjustiss-apple, line 1150: Do we want to carry `trellis_loop_diagonal_st8_prepatch_baseline_avx2` forward?

I agree it's a useful comparison point to prove the usefulness of the new approach, but I'm not sure it's worth preserving. Are you thinking we might need to revert back to the old TCQ approach in the future?

jjustiss-apple (approved):
LGTM, just a minor unit test maintenance question.

test/trellis_test.cc
chengchen-google: I don't think we need to revert back to the old TCQ approach.
I don't like the redundant code either. But since it is in the unit test only, it might be OK.

urvangjoshi:
@chengchen-google :  This is the unit test that is failing:

```
[ RUN      ] AVX2/TcqLoopDiagonalSt8Test.RandomValues/0
../../test/trellis_test.cc:1117: Failure
Expected equality of these values:
  rate_ref
    Which is: -1099013650
  rate_tst
    Which is: 193634

[  FAILED  ] AVX2/TcqLoopDiagonalSt8Test.RandomValues/0, where GetParam() = bit_depth:0 function:0x59d01f66a390 function:0x59d01fafb590 (0 ms)
[----------] 1 test from AVX2/TcqLoopDiagonalSt8Test (0 ms total)
```

For reference, this section of wiki page explains how to find the failing test: https://github.com/AOMediaCodec/avm/wiki/Reproducing-CI-Test-Failures-Locally#finding-failing-tests

PR #5331, opened 2026-08-26
DRAFT: Rebase av2enc on main

yeqing-wu:
Please don't squash — we need to preserve the commit history, so rebase only.


yeqing-wu:
Hi @urvangjoshi, @leolzhao, and @yunqingwang1,

This one is target for av2-enc branch. I've rebased the entire av2-enc branch on top of the main branch.

Meanwhile, I'm running the speed 0/1/4 tests on the rebased branch to confirm nothing is broken, and the results would look as expected.

Feel free to review in parallel and let me know if anything needs changing.

`Notes: Please don't squash — we need to preserve the commit history, so rebase only.`

Thanks
Yeqing

PR #5335, opened 2026-08-27
Add early termination to warp MV refinement

deepa-kg:
Added a speed feature to terminate warp MV
refinement iterations when the improvement
over the previous iteration is below 5%.

Enabled for speed >= 3.

Results for RA CTC, A1 17 frames, A2 33 frames,
speed 3: (Anchor: 735d104)

```
+-----+------+-------+-----+-------+---------------+
|Class|  Y   |  U    |   V  | YUV  |EncInstCount(%)|
+-----+------+-------+------+------+---------------+
| A1  | 0.04 | -0.28 | 0.05 | 0.02 | 97.9          |
| A2  | 0.04 | -0.21 | 0.10 | 0.02 | 98.0          |
+-----+------+-------+------+------+---------------+
```
STATS_CHANGED for speed >= 3

### DIFF
--- a/av2/encoder/mcomp.c
+++ b/av2/encoder/mcomp.c
@@ -4719,13 +4719,25 @@ static const warp_search_config warp_search_info[WARP_SEARCH_METHODS] = {
   },
 };
 
+// Checks if the warped motion vector refinement search can be terminated early.
+static AVM_INLINE bool early_terminate_refine_warped_mv(
+    uint64_t best_rd, uint64_t best_rd_prev, bool center_best_so_far,
+    bool do_early_terminate) {
+  if (center_best_so_far) return true;
+
+  if (do_early_terminate && (best_rd > 0.95 * best_rd_prev)) return true;
+
+  return false;
+}
+
 // Refines MV in a small range
 unsigned int av2_refine_warped_mv(MACROBLOCKD *xd, const AV2_COMMON *const cm,
                                   const SUBPEL_MOTION_SEARCH_PARAMS *ms_params,
                                   BLOCK_SIZE bsize, const int *pts0,
                                   const int *pts_inref0, int total_samples,
                                   int8_t ref, WARP_SEARCH_METHOD search_method,
-                                  int num_iterations) {
+                                  int num_iterations,
+                                  bool warp_mv_refine_early_term) {
   MB_MODE_INFO *mbmi = xd->mi[0];
 
   const MV *neighbors = warp_search_info[search_method].neighbors;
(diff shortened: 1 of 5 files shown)


deepa-kg:
@yunqingwang1 @urvangjoshi Please review
/cc @vikasprasad10 @harishdm @ranjit-tulabandu @cherma-a @RemyaPrakasan @ram-mohan

yunqingwang1 (approved):
Good tradeoff!

av2/encoder/mcomp.c
```diff
@@ -4719,13 +4719,25 @@ static const warp_search_config warp_search_info[WARP_SEARCH_METHODS] = {
...
+                                              uint64_t best_rd_prev,
+                                              int best_idx,
+                                              bool warp_mv_refine_early_term) {
+  if (best_idx == -1) return true;
+
+  if (warp_mv_refine_early_term && (best_rd > 0.95 * best_rd_prev)) return true;
+
+  return false;
+}
+
```
yeqing-wu, line 4732: Since you’ve defined this inline function, could you please help refactor the code in `av2_pick_warp_delta() ` to eliminate the duplicated code and maintain a consistent design. Additionally, could we rename the function from `early_terminate_refine()` to `early_terminate_refine_warped_mv() ` since it’s specifically designed for the warp mode, without the warp information, it would be more like for the refinemv tool. What are your thoughts on this?

```
      const int early_terminate_refine =
          center_best_so_far ||
          (early_term_warp_delta_refine && best_rd > 0.95 * best_rd_prev);

      if (early_terminate_refine) {
        break;
      }
```

av2/encoder/speed_features.h
```diff
@@ -582,6 +582,9 @@ typedef struct MV_SPEED_FEATURES {
   // Maximum number of iterations in WARP_CAUSAL refinement search
   int warp_search_iters;
 
+  // Early terminate warp MV refinement when improvement becomes small.
```
yeqing-wu: Small is a little bit vague here, should we make it be more clearly and which speed preset would enable this?
// Early-terminate warp MV refinement when the RD improvement over the
// previous iteration is less than ~5%. Enabled for speed >= 3.

yeqing-wu (commented):
Thanks for your efforts. Overall, the code looks good to me. Only two minor suggestions.

av2/encoder/mcomp.c
deepa-kg: Done

av2/encoder/speed_features.h
deepa-kg: Done

deepa-kg:
@yeqing-wu Please review.

yeqing-wu (approved):

PR #5337, opened 2026-08-27
Pickrst: Size restoration line buffers to frame width

ram-mohan:
- Replace the RESTORATION_LINEBUFFER_WIDTH based static buffers, sized for a max supported picture with buffers allocated per plane from the actual frame width.
- Replace per RU alloc and free in favor of per frame allocation

no stats changed


jianle973 (approved):
LGTM.

PR #5339, opened 2026-08-27
Fix out-of-range operating_points_count

xinzhao-apple:
The advertised 1-31 range is not aligned with the spec: ops_cnt is an f(3) field, so MAX_OPS_COUNT (7) is the correct upper bound.


SMPaluri (approved):
Thanks Xin

PR #5340, opened 2026-08-27
Bias towards large partition block

yunqingwang1:
Bias towards large partition block is enabled for high QP area in low complexity decode (LC) mode.

Only enabled for >=2160p now, CTC performance (65 frames, speed 1): baseline commit: 5d628d84
```
Test_Set  PSNR-YUV   SSIM    VMAF    ENC_TIME  DEC_TIME
  A1        0.04%   -0.24%  -0.02%    97%        96%
```
STATS_CHANGED only for LC mode

### DIFF
--- a/av2/encoder/partition_search.c
+++ b/av2/encoder/partition_search.c
@@ -52,6 +52,64 @@
 #include "av2/encoder/partition_ml.h"
 #endif
 
+static double get_partition_size_multiplier(BLOCK_SIZE bsize) {
+  switch (bsize) {
+    case BLOCK_128X128:
+    case BLOCK_128X64:
+    case BLOCK_64X128: return 1.0;
+    case BLOCK_64X64:
+    case BLOCK_64X32:
+    case BLOCK_32X64: return 1.005;  // 0.5% penalty
+    case BLOCK_32X32:
+    case BLOCK_32X16:
+    case BLOCK_16X32:
+    case BLOCK_16X64:
+    case BLOCK_64X16:
+    case BLOCK_8X64:
+    case BLOCK_64X8: return 1.01;  // 1% penalty
+    case BLOCK_16X16:
+    case BLOCK_16X8:
+    case BLOCK_8X16:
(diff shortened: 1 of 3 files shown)


av2/encoder/partition_search.c
```diff
@@ -5952,6 +5919,10 @@ bool av2_rd_pick_partition(
   }
 
   // Store the final rd cost
+  if (cpi->sf.lc_sf.enable_partition_size_bias &&
+      cm->current_frame.frame_type == INTER_FRAME)
```
yeqing-wu: can we put these into an inline function since it already has two places use the same code. I imagine it would be used in more places in the future?

```
cpi->sf.lc_sf.enable_partition_size_bias &&
      cm->current_frame.frame_type == INTER_FRAME
```

av2/encoder/partition_search.c
```diff
@@ -52,6 +52,50 @@
...
+}
+
+static void apply_partition_size_bias(RD_STATS *rdc, BLOCK_SIZE bsize,
+                                      bool found_partition) {
+  if (found_partition && rdc->rdcost < INT64_MAX && rdc->rdcost >= 0) {
+    const double mult = get_partition_size_multiplier(bsize);
+    if (mult != 1.0) {
+      rdc->rate = (int)(rdc->rate * mult);
+      rdc->dist = (int64_t)(rdc->dist * mult);
+      rdc->rdcost = (int64_t)(rdc->rdcost * mult);
```
yeqing-wu: can we guarantee it would not cause the overflow issue as rdcost multiplies a value that is larger than 1?

av2/encoder/partition_search.c
yunqingwang1: Done

av2/encoder/partition_search.c
yunqingwang1: Done -thanks.

yeqing-wu (approved):

PR #5345, opened 2026-08-28
Mark temporal point info metadata as suffix metadata

xinzhao-apple:
Temporal point info is prefix metadata (is_suffix == 0), but it was written after the tile group OBUs. The decoder treats a prefix metadata OBU that follows a coded frame as the start of the next frame unit, so --timing-info=model produced a stream that failed to decode with "Unspecified internal error" on a nonexistent trailing frame.

Fix (Encoder only): Set is_suffix = 1 to match where the OBU is written.

### DIFF
--- a/av2/encoder/bitstream.c
+++ b/av2/encoder/bitstream.c
@@ -6667,7 +6667,7 @@ static size_t write_temporal_point_info_metadata(AV2_COMP *const cpi,
   }
 
   // Set up metadata fields for individual OBU_METADATA
-  metadata->is_suffix = 0;
+  metadata->is_suffix = 1;
   metadata->cancel_flag = 0;
   metadata->persistence_idc = AVM_NO_PERSISTENCE;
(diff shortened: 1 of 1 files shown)


SMPaluri (approved):

SMPaluri:
Thanks Xin, I think it could also just change is_suffix = 1 so that it is appended at the end. But what you have is fine. I meant to add an encoder cli to set is_suffix for all the metadata.

xinzhao-apple:
> Thanks Xin, I think it could also just change is_suffix = 1 so that it is appended at the end. But what you have is fine. I meant to add an encoder cli to set is_suffix for all the metadata.

Thanks Seethal for the review and approval, I've updated the PR to just is_suffix = 1. Much smaller change and the decoded output is identical. Agreed that an encoder CLI (in a separate commit) to set is_suffix for all metadata is the more systematic fix.

Issue #5347, opened 2026-08-31
Encoder writes undecodable stream when an MFH OBU and frame resize are combined


xinzhao-apple:
### What version / commit were you testing with?

v1.0.0-101-g885b0a892

### What steps will reproduce the problem?

1. Encode with MFH OBU signaling on and a resize that gives the key frame and the inter
   frames different coded sizes:

   ```sh
   ./avmenc --width=160 --height=172 --i420 --input-bit-depth=8 --bit-depth=8 \
            --cpu-used=4 --passes=1 --threads=1 --limit=3 --ivf \
            --resize-mode=1 --resize-denominator=8 --resize-kf-denominator=16 \
            --enable-mfh-obu-signaling=1 -o out.ivf noise_160x172.yuv
   ```
2. Decode it with the reference decoder:

   ```sh
   ./avmdec -o out.yuv out.ivf --rawvideo
   ```

### Expected behavior

A stream just produced by the reference encoder decodes cleanly with the reference decoder.

### Actual behavior

The encoder succeeds (exit 0) but the decoder fails on the first inter frame (exit 1):

```
Warning: Failed to decode frame 2: Corrupt frame detected
Warning: Additional information: Invalid value for EOB position token
```

PR #5349, opened 2026-08-31
Fix inferred frame size when an MFH OBU and frame resize are both in use

xinzhao-apple:
This commit fixes a mismatch between encoder and decoder for inferring the frame size when both an MFH OBU and frame resize are in use.

When an MFH OBU is present and frame_size_override_flag is 0, the decoder infers the MFH frame size, instead of using max_frame_width/max_frame_height in sequence header. The encoder compared only against max_frame_\*, so under RPR a frame that matched max_frame_\* but differed from the MFH size was written with no size and decoded at the wrong dimensions, desyncing the parse. This commit adds comparison against the size the decoder actually infers.

Fix issue #5347.

### DIFF
--- a/av2/encoder/bitstream.c
+++ b/av2/encoder/bitstream.c
@@ -5296,10 +5296,19 @@ static AVM_INLINE void write_uncompressed_header(
             "Bridge frame frame_number is not equal to ref_buf order_hint");
       }
     } else {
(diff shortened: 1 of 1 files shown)


SMPaluri (approved):

urvangjoshi (approved):
Pls run clang-format.

PR #5365, opened 2026-09-02
Remove extra dc skip guard from the final encoding stage

jingninghan:
The related functions were never called, as the mode search stage would ensure that if an intra block has eob is 1, its tx type must be 2D-DCT.

Tested in speed 1 and 4 for a5, a4, and a2, no coding stats change.

### DIFF
--- a/av2/encoder/encodemb.c
+++ b/av2/encoder/encodemb.c
@@ -1447,12 +1447,6 @@ void av2_encode_block_intra(int plane, int block, int blk_row, int blk_col,
                             (!frame_is_intra_only(cm) &&
                              (INTRA_BLOCK_OPT_TYPE == TRELLIS_OPT ||
                               INTRA_BLOCK_OPT_TYPE == TRELLIS_DROPOUT_OPT));
-    const bool do_dropout = (frame_is_intra_only(cm) &&
-                             (KEY_BLOCK_OPT_TYPE == DROPOUT_OPT ||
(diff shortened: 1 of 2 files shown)


av2/encoder/encodemb.c
```diff
@@ -1479,41 +1473,8 @@ void av2_encode_block_intra(int plane, int block, int blk_row, int blk_col,
...
-        else
-          av2_optimize_b(args->cpi, x, plane, block, tx_size, tx_type,
-                         CCTX_NONE, &txb_ctx, &dummy_rate_cost);
-      }
-      if (do_dropout && !fsc_mode && !enable_parity_hiding) {
-        av2_dropout_qcoeff(x, plane, block, tx_size, tx_type,
-                           cm->quant_params.base_qindex);
-      }
-    }
+    if (*eob == 1 && plane == 0) assert(tx_type == DCT_DCT);
```
yeqing-wu: How about we improve this one so that we can avoid the condition check if it’s in release mode?
`assert(!(*eob == 1 && plane == 0) || tx_type == DCT_DCT);`

urvangjoshi: +1.   You can even use the `IMPLIES` macro we have for this:
```
assert(IMPLIES(condition, statement));
```

jingninghan: ack

I will put this on hold for now as unit test ran into failures.

jingninghan: done

PR #5369, opened 2026-09-04
Introduce dedicated types for primary/secondary transform representation

arnab-dey-12:
@yunqingwang1, @urvangjoshi Please Review
/cc @vikasprasad10 @harishdm @ranjit-tulabandu @cherma-a @deepa-kg @RemyaPrakasan @ram-mohan @mudassir-sg 

AV2 introduces a secondary transform type, but reuses AV1's `TX_TYPE` (an enum) representing primary transform types,  to represent every transform-related value: the primary type, the secondary set, the secondary type, and the packed representation. Overloading one enum across four distinct meanings creates readability issues and is prone to mistakes:
- **Ambiguous contents**. A `TX_TYPE` $x$ could hold the packed value, the primary type, the secondary set, or the secondary type. Determining which requires tracing back to where it was assigned.
- **Misleading comparisons**. `tx_type != DCT_DCT` reads as "the primary type is not `DCT_DCT`", but since `tx_type` is actually the packed representation, the check really asks whether any of the three fields is set.
- **No compile-time safety**. The packed value and the primary type share the same `TX_TYPE` type, so passing one where the other is expected isn't a compile error,  it silently produces the wrong value at runtime. For example, in tx_search.c, `primary_tx_type` and `packed_tx_type` are both plain `TX_TYPE` in the same function, swapping them would compile cleanly while corrupting the transform type used downstream.

This PR introduces dedicated data structures so each concept is explicit and distinct:
- `PRIM_TX_TYPE` — the primary transform type
- `SEC_TX_SET` — the secondary transform set
- `SEC_TX_TYPE` — the secondary transform type

It also refactors the existing `TX_TYPE` enum into a bitfield union, and updates call sites to use the new types where appropriate. Together, these changes make the distinction between primary type, secondary set, secondary type, and packed representation explicit and make the code easier to read and maintain.

We have verified that this PR is bit-exact for speeds 0 to 3 across RA, LD, and AI configurations on a few test clips.

*Note* : Each commit is independent, self-contained change and has been verified to be bit-exact. Please do not squash.


yunqingwang1 (approved):
This refactoring looks good. Thanks.

av2/encoder/ethread.c
```diff
@@ -33,7 +33,7 @@ static AVM_INLINE void accumulate_rd_opt(ThreadData *td, ThreadData *td_t) {
   td->rd_counts.skip_mode_used_flag |= td_t->rd_counts.skip_mode_used_flag;
 
   for (int i = 0; i < TX_SIZES_ALL; i++) {
-    for (int j = 0; j < TX_TYPES; j++)
+    for (int j = 0; j < PRIM_TX_TYPES; j++)
```
leolzhao: Suggest to use PRIMARY_TX_TYPES for easy understanding. Same for the name of variables below.

av2/encoder/hybrid_fwd_txfm.c
```diff
@@ -1056,7 +1056,7 @@ void av2_fwd_cross_chroma_tx_block_c(tran_low_t *coeff_c1, tran_low_t *coeff_c2,
 
 void av2_fwd_stxfm(tran_low_t *coeff, TxfmParam *txfm_param,
                    int64_t *sec_tx_sse) {
-  const TX_TYPE stx_type = txfm_param->sec_tx_type;
+  const SEC_TX_TYPE stx_type = txfm_param->sec_tx_type;
```
leolzhao, line 1059: For this one, I'm fine with SEC_TX_TYPE as there is no confusion here. If you would like to change it to SECONDARY_TX_TYPE, I'm also ok with it.

av2/encoder/ethread.c
arnab-dey-12: Done

av2/encoder/hybrid_fwd_txfm.c
arnab-dey-12: Acknowledged.

leolzhao (approved):

PR #5372, opened 2026-09-05
Fix mode cost for palette search in inter frames

ram-mohan:
mbmode_cost is not initialized and used for cost computation during palette mode search in inter frame. This is replaced with y_primary_flag_cost + y_mode_idx_cost similar to iframe palette-search.

STATS_CHANGED

### DIFF
--- a/av2/encoder/block.h
+++ b/av2/encoder/block.h
@@ -810,8 +810,6 @@ typedef struct {
    * \name Intra Costs: General
    ****************************************************************************/
   /**@{*/
-  //! Luma mode cost for inter frame.
-  int mbmode_cost[BLOCK_SIZE_GROUPS][INTRA_MODES];
   //! intra_dip_cost
   int intra_dip_cost[DIP_CTXS][2];
   //! intra_dip_mode_cost
--- a/av2/encoder/intra_mode_search.c
+++ b/av2/encoder/intra_mode_search.c
@@ -854,9 +854,9 @@ int av2_search_palette_mode(IntraModeSearchState *intra_search_state,
   MB_MODE_INFO best_mbmi_palette = *mbmi;
   uint8_t best_blk_skip[MAX_MIB_SIZE * MAX_MIB_SIZE];
   TX_TYPE best_tx_type_map[MAX_MIB_SIZE * MAX_MIB_SIZE];
-  const ModeCosts *mode_costs = &x->mode_costs;
-  const int *const intra_mode_cost =
-      mode_costs->mbmode_cost[size_group_lookup[bsize]];
+  const int context = get_y_mode_idx_ctx(xd);
+  int mode_cost = x->mode_costs.y_primary_flag_cost[DC_PRED];
+  mode_cost += x->mode_costs.y_mode_idx_costs[context][DC_PRED];
   const int rows = block_size_high[bsize];
   const int cols = block_size_wide[bsize];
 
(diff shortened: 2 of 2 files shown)


ram-mohan:
@yunqingwang1 @urvangjoshi when you get a chance can you please review

yunqingwang1:
Do you have the CTC test results before/after this fix (especially for the sets that use palette mode)?

This part of code is from AV1, and if there is a bug, we need to fix both. Could you explain more on how mbmode_cost is not initialized and when this happens? Thanks.

ram-mohan:
@yunqingwang1, This issue is not present in av1. The mbmode_cost field is initialized in av1_fill_mode_rates, but in av2 this is field is unused. The inter-frame palette search was therefore using a never-written value as dc_mode_cost.
This is now modified to 
```c
int mode_cost  = mode_costs->y_primary_flag_cost[DC_PRED];
    mode_cost += mode_costs->y_mode_idx_costs[context][DC_PRED];
```
The mode cost computations shared here are similar to the ones in intra_mode_search.c. This is also cross verified against the decoder syntax at decodemv.c:1510-1517 (y_mode_set_cdf, then y_mode_idx_cdf[context]) and reordered_y_mode[0] == DC_PRED.

| Speed | Class | Y | Cb | Cr | wAvg | Enc% |
|---|---|---|---|---|---|---|
| 0 | B2 | -0.0414 | -0.4589 | 0.0818 | -0.0532 | 100.006 |
| 1 | B2 | 0.2856 | 0.3039 | 0.138 | 0.2804 | 100.066 |
| 2 | B2 | -0.1121 | -0.1039 | 0.5986 | -0.0834 | 100.036 |

This change has caused a quality drop in speed 1. A drop of 1.4% is seen for the asset MissionControlClip1_1920x1080_60fps_10bit_420_0450_0579.y4m and 1.24% for asset Spreadsheet_1920x1080_30fps_8bit_420_130f.y4m. We are currently looking in to this.

yunqingwang1:
Thanks Ram for the explanation. Do you think this can be handled similarly like av1 and gets initialized in av2_fill_mode_rates()? Also, since you are looking into the quality loss issue (probably some speed features don't work well with B2, and we saw it before), maybe we can hold on this PR for now?

ram-mohan:
@yunqingwang1, i apologize for not communicating my intent behind the fix correctly.
 
In av1, luma intra mode cost uses two tables: mbmode_cost (in inter frames) and y_mode_costs (in intra frames), each indexed by the prediction mode. In av2, the luma intra mode is signalled differently. It is coded as mode_set_index, mode_set_low, mode_idx - mode_set_low. av2_fill_mode_rates() fills the necessary cost tables for these. It fills y_primary_flag_cost for mode_set_index, y_mode_idx_costs for mode_set_low, y_mode_idx_offset_costs for mode_idx - mode_set_low. intra_mode_search.c of av2 uses these tables in its luma mode cost computation. 

The mbmode_cost table is which is part of av1 ModeCosts carried to av2, is no longer relevant. The code that initialized the mbmode_cost in av1 is removed in av2 but the field has remained. The only code still using it is palette mode computation. This is stale and needs to be updated according to the av2 requirements. This change handles this. Also, it removes the field mbmode_cost from the context.

yunqingwang1:
Hi Ram, thanks for the explanation. This makes good sense now. The speed 0 B2 coding performance is neutral with this fix. This result can be enough to verify this fix.

yunqingwang1 (approved):

urvangjoshi:
@ram-mohan @yunqingwang1  :  the fix by itself makes sense, so I'll go ahead and merge it.

For B2 set drop in speed 1,  you can test separately and fix it in separate PR if there's an issue found later.
(Sometimes B2 set can be sensitive to small changes).

PR #5375, opened 2026-09-08
Apply 8x8 partition floor at 1080p from speed 1

yeqing-wu:
The encoder already sets default_min_partition_size to 8x8 at 4k
and larger for all speeds. Extend that floor to 1080p and larger
from speed 1, and drop the now-duplicate 1080p check under
speed >= 6.

Thread the current speed through av2_alloc_context_buffers() to
enc_set_mb_mi() so mi_alloc_bsize tracks the same partition floor
without stashing speed on mi_params. RT keeps its pre-existing
mi_alloc_bsize behavior (always 8x8 at 4k) and will be revisited
separately.

Anchor: commit 4ced699
CTC: FG16 CTC (33 frames, class A1 and A2, RA)

```
1) Speed 1 (cpu-used=1)
  | A1         |  0.00 |  0.00 |  0.00 |  0.00 |  100 |  100 |
  | A2         |  0.14 |  0.32 |  0.85 |  0.17 |   93 |   99 |
  | Avg w/o B2 |  0.10 |  0.22 |  0.60 |  0.12 |   95 |   99 |

2) Speed 2 (cpu-used=2)
  | A1         |  0.00 |  0.00 |  0.00 |  0.00 |  100 |  100 |
  | A2         |  0.04 |  0.04 |  0.30 |  0.05 |   97 |  100 |
  | Avg w/o B2 |  0.03 |  0.03 |  0.21 |  0.03 |   98 |  100 |

3) Speed 3 (cpu-used=3)
  | A1         |  0.00 |  0.00 |  0.00 |  0.00 |  100 |  100 |
  | A2         |  0.09 |  0.30 |  0.11 |  0.10 |   96 |  100 |
  | Avg w/o B2 |  0.06 |  0.21 |  0.08 |  0.07 |   97 |  100 |

4) Speed 4 (cpu-used=4)
  | A1         |  0.02 |  0.02 |  0.00 |  0.02 |  100 |  100 |
  | A2         |  0.20 |  0.27 |  0.00 |  0.20 |   94 |  100 |
  | Avg w/o B2 |  0.15 |  0.19 |  0.00 |  0.14 |   96 |  100 |
```

STATS_CHANGED

### DIFF
--- a/av2/common/av2_common_int.h
+++ b/av2/common/av2_common_int.h
@@ -1613,9 +1613,8 @@ struct CommonModeInfoParams {
   int mi_alloc_stride;
   /*!
    * The minimum block size that each element in 'mi_alloc' can correspond to.
-   * For decoder, this is always BLOCK_4X4.
-   * For encoder, this is currently set to BLOCK_4X4 for resolution < 4k,
-   * and BLOCK_8X8 for resolution >= 4k.
+   * For decoder, always BLOCK_4X4. For encoder, BLOCK_8X8 at 4k, and at 1080p
+   * from speed 1; BLOCK_4X4 otherwise. Computed in 'enc_set_mb_mi()'.
    */
   BLOCK_SIZE mi_alloc_bsize;
 
@@ -1702,9 +1701,10 @@ struct CommonModeInfoParams {
    * \param[in,out]   mi_params   object containing common mode info parameters
    * \param           width       frame width
    * \param           height      frame height
+   * \param           speed       encoder speed level (ignored by the decoder)
    */
   void (*set_mb_mi)(struct CommonModeInfoParams *mi_params, int width,
-                    int height);
+                    int height, int speed);
   /**@}*/
 };
 
--- a/av2/common/alloccommon.c
+++ b/av2/common/alloccommon.c
@@ -600,9 +600,10 @@ int av2_alloc_superblock_info_buffers(AV2_COMMON *cm) {
   return alloc_sbi(sbi_params);
 }
 
-int av2_alloc_context_buffers(AV2_COMMON *cm, int width, int height) {
+int av2_alloc_context_buffers(AV2_COMMON *cm, int width, int height,
+                              int speed) {
   CommonModeInfoParams *const mi_params = &cm->mi_params;
-  mi_params->set_mb_mi(mi_params, width, height);
+  mi_params->set_mb_mi(mi_params, width, height, speed);
   if (alloc_mi(mi_params, cm, height)) goto fail;
 
   if (av2_alloc_superblock_info_buffers(cm)) goto fail;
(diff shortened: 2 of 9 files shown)


yunqingwang1 (approved):
Looks good.

av2/common/av2_common_int.h
```diff
@@ -1629,12 +1629,25 @@ struct CommonModeInfoParams {
...
-   * For decoder, this is always BLOCK_4X4.
-   * For encoder, this is currently set to BLOCK_4X4 for resolution < 4k,
-   * and BLOCK_8X8 for resolution >= 4k.
+   * For decoder, always BLOCK_4X4. For encoder, BLOCK_8X8 when the partition
+   * floor is guaranteed to stop there (GOOD at 4k, GOOD at 1080p with
+   * 'enc_speed' >= 1); BLOCK_4X4 otherwise. Computed in 'enc_set_mb_mi()'.
    */
   BLOCK_SIZE mi_alloc_bsize;
 
+  /*!
```
urvangjoshi: Adding these two values in `mi_params`  seems risky -- as things may easily fall out-of-sync if we aren't careful during code changes.

Pls use the original logic as is -- just changing which speed this setting belongs to.

For RT issue,  we can change that one in a separate PR --  by e.g. adding a `set_rt_speed_feature_framesize_dependent` function to ensure there is no inconsistency.
FYI @jianj-g

urvangjoshi (changes requested):

av2/common/av2_common_int.h
yeqing-wu: @urvangjoshi, sounds good to me, we just need to fix the issue in RT later.

leolzhao (approved):
Good tradeoff!

av2/common/av2_common_int.h
yeqing-wu: @urvangjoshi, I’ve modified the code to eliminate those two items. However, we still need to pass the speed to enc_set_mb_mi() and av2_enc_partition_floor(). The issue arises because the code uses the common function pointer API set_mb_mi() for both encoder enc_set_mb_mi() and stat_stage_set_mb_mi() and decoder dec_set_mb_mi(). Consequently, we must pass an unused speed to the decoder API and the upper-level common functions that call dec_set_mb_mi().

Another solution is to leave mi_alloc_bsize unchanged for 1080p with a speed of at least 1, still 4x4. This decision could lead to inconsistent design for mi grid size compared to sf->part_sf.default_min_partition_size.

Which one do you prefer?

yeqing-wu: @urvangjoshi , any feedback for this?

urvangjoshi: Thanks @yeqing-wu !  I understand the reason for additional arg to `set_mb_mi`  and agree with this solution.

urvangjoshi (approved):

PR #5376, opened 2026-09-09
Transform search result cache speed feature for speed 1

gjin-rgb:
Within a frame, remember the winning transform type of a luma transform block, keyed on its residual and on every context value that can change which candidate wins. On a hit the primary transform type mask and the intra secondary transform search are narrowed to that winner, which is still evaluated by the normal path, so only the search breadth shrinks.

The table is 64K entries of 16 bytes, embedded in TxfmSearchInfo, so it is per-thread and needs no allocation. Each entry carries the frame it was produced in, so entries from earlier frames are ignored and recycled instead of cleared.

Enabled for speed 1 and only for sequence quantizers below 210 + MAXQ_OFFSET * (bit_depth - 8): at coarser quantizers the work this prunes has already been taken by the DC-only and skip predictors and the transform-domain RD estimate, while the residual hash costs the same, so there is no speed-up left to win there.

Results for RA CTC, A1 17 frames, A2 33 frames, speed 1: (Anchor: ee0b4ba10)

```
+-----+------+-------+-------+------+---------------+
|Class|  Y   |   U   |   V   | YUV  |Enc wall clock |
+-----+------+-------+-------+------+---------------+
| A1  | 0.03 |  0.01 |  0.03 | 0.03 | 97.2          |
| A2  | 0.02 |  0.18 |  0.13 | 0.03 | 97.9          |
+-----+------+-------+-------+------+---------------+
```

STATS_CHANGED for speed 1

### DIFF
--- a/av2/encoder/tx_cache.h
+++ b/av2/encoder/tx_cache.h
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2021, Alliance for Open Media. All rights reserved
+ *
+ * This source code is subject to the terms of the BSD 3-Clause Clear License
+ * and the Alliance for Open Media Patent License 1.0. If the BSD 3-Clause Clear
+ * License was not distributed with this source code in the LICENSE file, you
+ * can obtain it at aomedia.org/license/software-license/bsd-3-c-c/.  If the
+ * Alliance for Open Media Patent License 1.0 was not distributed with this
+ * source code in the PATENTS file, you can obtain it at
+ * aomedia.org/license/patent-license/.
+ */
+
+// Transform search result cache: storage only.  An entry holds the transform
+// type that won for a key, plus the frame it was produced in so that entries
+// from earlier frames are ignored and recycled instead of cleared.  The table
+// is embedded in TxfmSearchInfo, so it is per-thread and needs no allocation.
+
+#ifndef AVM_AV2_ENCODER_TX_CACHE_H_
+#define AVM_AV2_ENCODER_TX_CACHE_H_
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define TX_CACHE_BITS 16  // 64K entries, 1 MiB
+#define TX_CACHE_SIZE (1u << TX_CACHE_BITS)
+#define TX_CACHE_MASK (TX_CACHE_SIZE - 1u)
+#define TX_CACHE_PROBE 8  // linear probe depth
+
+/*! Cache slot: a fully mixed key, the transform type that won for it, and the
+ *  frame it belongs to (frame_number + 1; 0 means the slot was never written).
+ */
+typedef struct {
+  uint64_t key;
+  uint16_t winner;  // full tx_type: primary | secondary | secondary set
+  uint32_t tag;
+} TxCacheEntry;
+
+/*! Transform search result cache. Entries are scoped to the frame they were
+ *  produced in, so the table is never cleared.
+ */
+typedef struct {
+  TxCacheEntry entries[TX_CACHE_SIZE];
+} TxCache;
+
+#ifdef __cplusplus
+}  // extern "C"
+#endif
+
+#endif  // AVM_AV2_ENCODER_TX_CACHE_H_
--- a/av2/encoder/tx_cache.h
+++ b/av2/encoder/tx_cache.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, Alliance for Open Media. All rights reserved
+ * Copyright (c) 2026, Alliance for Open Media. All rights reserved
  *
  * This source code is subject to the terms of the BSD 3-Clause Clear License
  * and the Alliance for Open Media Patent License 1.0. If the BSD 3-Clause Clear
@@ -10,10 +10,10 @@
  * aomedia.org/license/patent-license/.
  */
 
-// Transform search result cache: storage only.  An entry holds the transform
-// type that won for a key, plus the frame it was produced in so that entries
-// from earlier frames are ignored and recycled instead of cleared.  The table
-// is embedded in TxfmSearchInfo, so it is per-thread and needs no allocation.
+// Transform search result cache.  An entry holds the transform type that won
+// for a key, plus the frame it was produced in so that entries from earlier
+// frames are ignored and recycled instead of cleared.  The table is embedded
+// in TxfmSearchInfo, so it is per-thread and needs no allocation.
 
 #ifndef AVM_AV2_ENCODER_TX_CACHE_H_
 #define AVM_AV2_ENCODER_TX_CACHE_H_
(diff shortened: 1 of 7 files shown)


leolzhao (approved):

leolzhao:
Please fix the unit test failures. @gjin-rgb

av2/encoder/tx_cache.h
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) 2021, Alliance for Open Media. All rights reserved
```
urvangjoshi: 2026

av2/encoder/tx_search.c
```diff
@@ -22,6 +22,93 @@
 #include "av2/encoder/rdopt_utils.h"
 #include "av2/encoder/tx_prune_model_weights.h"
 #include "av2/encoder/tx_search.h"
+#include "av2/encoder/tx_cache.h"
+
+// Transform search result cache, see av2/encoder/tx_cache.h.
```
urvangjoshi: You can create a `tx_cache.c` and define these cache related functions there   (and declare them in `tx_cache.h` that you already created).

av2/encoder/tx_cache.h
```diff
@@ -0,0 +1,52 @@
...
+#define TX_CACHE_MASK (TX_CACHE_SIZE - 1u)
+#define TX_CACHE_PROBE 8  // linear probe depth
+
+/*! Cache slot: a fully mixed key, the transform type that won for it, and the
+ *  frame it belongs to (frame_number + 1; 0 means the slot was never written).
+ */
+typedef struct {
+  uint64_t key;
+  uint16_t winner;  // full tx_type: primary | secondary | secondary set
+  uint32_t tag;
```
urvangjoshi: Add a comment for `tag`

av2/encoder/tx_cache.h
```diff
@@ -0,0 +1,52 @@
...
+#define TX_CACHE_BITS 16  // 64K entries, 1 MiB
+#define TX_CACHE_SIZE (1u << TX_CACHE_BITS)
+#define TX_CACHE_MASK (TX_CACHE_SIZE - 1u)
+#define TX_CACHE_PROBE 8  // linear probe depth
+
+/*! Cache slot: a fully mixed key, the transform type that won for it, and the
+ *  frame it belongs to (frame_number + 1; 0 means the slot was never written).
+ */
+typedef struct {
+  uint64_t key;
```
urvangjoshi: Add a comment to explain what is used as the key here

av2/encoder/tx_search.c
```diff
@@ -22,6 +22,93 @@
...
+// Cached winning tx_type for this frame, or -1 on a miss. A never-written slot
+// (tag 0) ends the probe: no store for this frame can have walked past it.
+static AVM_INLINE int tx_cache_lookup(const TxCache *cache, uint64_t key,
+                                      uint32_t frame) {
+  const uint32_t tag = frame + 1;
+  const uint32_t idx = (uint32_t)key & TX_CACHE_MASK;
+  for (int p = 0; p < TX_CACHE_PROBE; ++p) {
+    const TxCacheEntry *e = &cache->entries[(idx + p) & TX_CACHE_MASK];
+    if (e->tag == 0) return -1;
+    if (e->tag == tag && e->key == key) return (int)e->winner;
```
urvangjoshi: I see a number of places doing signed / unsigned integer casts (here and elsewhere). Can you double check that these are all necessary?

av2/encoder/tx_search.c
```diff
@@ -2527,6 +2614,52 @@ static void search_tx_type(const AV2_COMP *cpi, MACROBLOCK *x, int plane,
...
+        tx_cache, tx_cache_key, (uint32_t)cm->current_frame.frame_number);
+    if (cached >= 0) {
+      tx_cache_hit = 1;
+      tx_cache_winner.packed_tx_type = (uint16_t)cached;
+      const uint16_t winner_bit = (uint16_t)(1u << tx_cache_winner.primary_tx);
+      // Keep DCT_DCT in the mask: the prune gates inside the loop can reject
+      // the winner, and a single-candidate mask would then leave the search
+      // with nothing evaluated.
+      if (allowed_tx_mask & winner_bit) {
+        allowed_tx_mask &= (uint16_t)(winner_bit | 1u);
```
urvangjoshi: Is the "| 1u"  part corresponding to the comment "Keep DCT_DCT in the mask...."?
It's not really obvious.

If that is correct, could you do something like this?
- Move the comment inside the `if` here
- Also, how about defining `const uint16_t dct_bit = (1u << DCT_DCT);`  to make this explicit and then use `dct_bit` in this line?

av2/encoder/tx_cache.h
gjin-rgb: Done

av2/encoder/tx_search.c
gjin-rgb: Done. Also added both tx_cache.c and tx_cache.h to av2/av2.cmake — the header was missing from the source .

av2/encoder/tx_search.c
gjin-rgb: Good catch — most were not. Removed unnecessary casts.

av2/encoder/tx_search.c
gjin-rgb: Correct, that was the DCT_DCT bit (DCT_DCT == 0) . I changed the line to make it more clear.

av2/encoder/tx_cache.h
gjin-rgb: Added

av2/encoder/tx_cache.h
gjin-rgb: Added

av2/encoder/tx_search.c
```diff
@@ -2527,6 +2528,52 @@ static void search_tx_type(const AV2_COMP *cpi, MACROBLOCK *x, int plane,
...
+        av2_get_ext_tx_set_type(tx_size, is_inter,
+                                cm->features.reduced_tx_set_used),
+        av2_use_qmatrix(&cm->quant_params, xd, mbmi->segment_id), x->rdmult);
+    const int cached = av2_tx_cache_lookup(tx_cache, tx_cache_key,
+                                           cm->current_frame.frame_number);
+    if (cached >= 0) {
+      tx_cache_hit = 1;
+      tx_cache_winner.packed_tx_type = (uint16_t)cached;
+      const uint16_t winner_bit = 1 << tx_cache_winner.primary_tx;
+      const uint16_t dct_bit = 1 << DCT_DCT;
```
urvangjoshi: This should be moved inside the `if` (and just after the comment)

urvangjoshi (approved):

av2/encoder/tx_search.c
gjin-rgb: yes

urvangjoshi:
@gjin-rgb : pls don't create merge commits.  Rebase and force push (`git push --force`) instead.

This will help us rebase and merge the PR ourselves once CI passes (as long as there are no conflicts)

PR #5380, opened 2026-09-09
Bias against CDEF

yunqingwang1:
This change is for low-complexity decode mode. Bias against CDEF during frame
level RD decision. Also disabled CDEF for skip blocks.

```
RA CTC test with "--enable-low-complexity-decode=2"
65 frames / speed 1 (base: ad9c4353)
 +-----+-------+--------+-------+----------+----------+
 |Class| WPSNR | SSIM-Y | VMAF  | Enc-time | Dec-time |
 +-----+----- -+-- -----+-------+----------+----------+
 | A1  | 0.40  | -0.18  | -0.36 |   99     |   93     |
 | A2  | 0.47  | -0.19  | -0.13 |   99     |   94     |
 +-----+-------+--------+-------+----------+----------+
```

STATS_CHANGED for "--enable-low-complexity-decode=2"

### DIFF
--- a/av2/encoder/encoder.c
+++ b/av2/encoder/encoder.c
@@ -3137,7 +3137,8 @@ static void cdef_restoration_frame(AV2_COMP *cpi, AV2_COMMON *cm,
 #if CONFIG_ENTROPY_STATS
                     &cpi->td,
 #endif  // CONFIG_ENTROPY_STATS
-                    cpi->sf.lpf_sf.cdef_pick_method, cpi->td.mb.rdmult);
+                    cpi->sf.lpf_sf.cdef_pick_method, cpi->td.mb.rdmult,
+                    cpi->sf.lc_sf.bias_against_cdef);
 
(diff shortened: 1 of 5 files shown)


yeqing-wu (approved):
LGTM

urvangjoshi:
Questions @yunqingwang1 :  
- Looks like this is for low-complexity mode?  Pls clarify this in the PR description and which level of LC it applies to.
- Do you know why there is a loss on PSNR but gain in VMAF?

yunqingwang1:
> Questions @yunqingwang1 :
> 
> * Looks like this is for low-complexity mode?  Pls clarify this in the PR description and which level of LC it applies to.

Yes, I added more details in commit message.

> * Do you know why there is a loss on PSNR but gain in VMAF?

Regarding the VMAF gain, from the ctc results, I can see not every clip benefits from enabling CDEF all the time. Less CDEF seems good for no-sharp videos.

PR #5385, opened 2026-09-10
Disable tx partition at speed 6

motongxu:
Tx partition is disabled for speed >= 6.

Anchor: https://github.com/AOMediaCodec/avm/commit/59be49b8ecb360f3e419acba3429a21b15059935
| A1      |  1.38% |  0.40% |  0.48% |  1.27% | 76.5%    | 101%     |
| A2      |  1.43% |  0.47% |  0.81% |  1.35% | 78.4%    | 100%     |

### DIFF
--- a/av2/encoder/speed_features.c
+++ b/av2/encoder/speed_features.c
@@ -689,6 +689,7 @@ static void set_good_speed_features_framesize_independent(
 
     sf->tx_sf.tx_type_search.winner_mode_tx_type_pruning = 2;
     sf->tx_sf.tx_type_search.prune_tx_type_est_rd = 0;
+    sf->tx_sf.enable_tx_partition = 0;
 
     sf->rd_sf.perform_coeff_opt = is_boosted_arf2_bwd_type ? 4 : 6;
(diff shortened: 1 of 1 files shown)


yunqingwang1 (approved):

yeqing-wu (approved):

av2/encoder/speed_features.c
```diff
@@ -689,6 +689,7 @@ static void set_good_speed_features_framesize_independent(
 
     sf->tx_sf.tx_type_search.winner_mode_tx_type_pruning = 2;
     sf->tx_sf.tx_type_search.prune_tx_type_est_rd = 0;
+    sf->tx_sf.enable_tx_partition = 0;
```
yeqing-wu: just one minor comment: enable_tx_partition data type is bool, thus it is better to use false instead of 0. The code is already not consistent, maybe okay just using 0.

motongxu: Thanks for the comment. It is changed to "false" in version 2.

PR #5386, opened 2026-09-10
Removed unused tmp pred buf and GDF line buffer memory

ram-mohan:
tmp_pred_bufs: collapse the [2] array to a single buffer (the second element has been unused since OBMC was removed) and drop the stale 2x size multiplier left over from the uint8_t -> uint16_t buffer type change.

GDFLineBuffers: allocate the save above/below line buffers to the actual frame width instead of RESTORATION_LINEBUFFER_WIDTH.

### DIFF
--- a/av2/common/av2_common_int.h
+++ b/av2/common/av2_common_int.h
@@ -433,12 +433,10 @@ typedef struct BufferPool {
  */
 typedef struct {
   uint16_t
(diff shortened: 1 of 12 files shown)


ram-mohan:
@yunqingwang1 @urvangjoshi when you get a chance can you please review

urvangjoshi (approved):

PR #5387, opened 2026-09-10
Enable the dc-only block predictor at speed 2 above a qindex threshold

gjin-rgb:
- Gate in av2_set_speed_features_qindex_dependent(): enable when base_qindex > 113 + MAXQ_OFFSET * (bit_depth - 8).
- New predict_dc_levels row { 2, 2, 0 } takes the dc-only outcome only.

  Class   BD-rate (PSNR wAvg)   Enc instr   Enc wall
  A1              +0.0657%         97.0%      97.8%
  A2              +0.0062%         98.4%      98.6%
Anchor (b42be4efa6af35ff1fca51a1ff1052aedd97322f)
STATS_CHANGED

### DIFF
--- a/av2/encoder/speed_features.c
+++ b/av2/encoder/speed_features.c
@@ -119,9 +119,10 @@ static unsigned int predict_skip_levels[3][MODE_EVAL_TYPES] = { { 0, 0, 0 },
 // Values indicate the aggressiveness of skip flag prediction.
 // 0 : no early DC block prediction
 // 1 : Early DC block prediction based on error variance
-static unsigned int predict_dc_levels[3][MODE_EVAL_TYPES] = { { 0, 0, 0 },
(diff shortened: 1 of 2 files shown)


jingninghan:
What's the rationale of picking QP = 113 as the threshold? What happens if taking out the QP dependency for a consistent behavior?

gjin-rgb:
> What's the rationale of picking QP = 113 as the threshold? What happens if taking out the QP dependency for a consistent behavior? 
@jingninghan 
113 is not a new constant — this function already uses 113 + qindex_offset as a qindex threshold in two other places (lines 1569 and 1593), so I reused it rather than introduce another magic number. I did compare the gated and ungated variants, but the difference is entirely at the low-QP end.

jingninghan (approved):

PR #5388, opened 2026-09-10
Deprecate check_primary_quant_all_zero

jingninghan:
The quantization function itself is not a major bottleneck. This check_primary_quant_all_zero() would increasingly return as false as the bit-rate goes higher, which makes it waste cpu cycles in the higher bit-rate range.

### DIFF
--- a/av2/encoder/tx_search.c
+++ b/av2/encoder/tx_search.c
@@ -2118,29 +2118,6 @@ static bool prune_sec_txfm_rd_eval(int64_t sec_tx_sse_to_be_coded,
   return false;
 }
 
-// Return 1 if primary FP-quant is guaranteed to produce eob = 0 across the
-// IST window, 0 otherwise. Uses an L-infinity scan of the post-primary
-// coefficients against the FP quantizer's kill threshold
-//   abs(coeff) < min(dequant[0], dequant[1]) >> (4 + log_scale)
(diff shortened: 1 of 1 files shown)


av2/encoder/tx_search.c
```diff
@@ -2613,30 +2590,6 @@ static void search_tx_type(const AV2_COMP *cpi, MACROBLOCK *x, int plane,
         }
         *coeffs_available = 1;
 
-        // Pre-IST quant-zero gate. Predict primary FP-quant eob = 0 from the
-        // post-primary coefficients and, if so, skip av2_quant / trellis /
-        // cost_coeffs and propagate the same eob_found / DCT_DCT handling as
-        // the post-quant eob == 1 gate below. Scoped to intra Y, stx == 0,
-        // non-DC-only, IST-enabled, non-QM blocks.
-        if (tx_sf->prune_intra_ist_stx_by_zero_eob && plane == PLANE_TYPE_Y &&
```
urvangjoshi, line 2675: Is `prune_intra_ist_stx_by_zero_eob` speed feature still used after this change?  If not, it could be cleaned up?

jingninghan: This speed feature flag is used later on.

urvangjoshi (approved):

PR #5393, opened 2026-09-11
Unrestricted partitioning depth for H partitions at speed 1

urvangjoshi:
Keep ext_recur_depth_level = 0, to allow unrestricted partitioning for H partitions.
Only affects speed 1 because H partitions are disabled at speed >= 2.

Speed 1 CTC 33 frames RA:
(Anchor: 4a8df58a7)

```
+---------+--------+-------------+
| Summary |  YUV   |   Enc-time  |
+---------+--------+-------------+
|   A1    | -0.15% |   103.8%   |
|   A2    | -0.22% |   107.0%   |
|Avg wo B2| -0.27% |   108.0%   |
+---------+--------+-------------+
```

STATS_CHANGED for speed 1


urvangjoshi:
Pls review. I have checked that tradeoff is good matching the expected ratio for speed 1

yunqingwang1:
@urvangjoshi thanks.
@leolzhao @yeqing-wu We didn't clearly define a tradeoff for quality improving. I feel that should be much lower than the optimization tradeoff (i.e. 35 for speed 1). This PR's A2 set tradeoff is 31.8 that is somewhat close to 35. What do you think?

yeqing-wu:
> @urvangjoshi thanks. @leolzhao @yeqing-wu We didn't clearly define a tradeoff for quality improving. I feel that should be much lower than the optimization tradeoff (i.e. 35 for speed 1). This PR's A2 set tradeoff is 31.8 that is somewhat close to 35. What do you think?

To me, the overall encoding time increase is a lot, with 8% being quite substantial. I think the ratio is approximately near to the bar. Is it possible to reduce the encoding time by half while accepting a certain level of coding loss?

urvangjoshi:
@yunqingwang1 and @yeqing-wu :  Thanks for the comments. 
I'd say we need to think in this way:  if a tool was already on at speed 1,  and a PR came in to disable it with a ratio of 25 for A2 and 31 for A3,  will we accept that PR?

_[Note:   to be precise,  the A2 ratio for this situation will be in fact `(1 - (100 / 107)) / 0.22 = 0.30`]_

My understanding that we won't accept this PR. Given that,  we should logically turn such a tool on.
What do you think?

On a separate note:  the trade-off  for this used to be worse,  but other speed-ups have already helped and brought the trade-off within the limits.

yunqingwang1:
@urvangjoshi I understand what you said. Just share my thoughts here. We probably want to pay more attention to the encoder speed since that is the primary goal for av2-enc. As you noticed, one speed feature's tradeoff can shift as other features are integrated. While reverting a feature for coding gains, it is preferred to do it conservatively (i.e. With a much better tradeoff than the current bar). As Yeqing mentioned, it would be great if the speed loss can be reduced. Maybe only allow unrestricted H partitioning for low pyramid level frames? Or make it a resolution dependent feature? Please correct me if I miss anything.

urvangjoshi:
> @urvangjoshi I understand what you said. Just share my thoughts here. We probably want to pay more attention to the encoder speed since that is the primary goal for av2-enc. As you noticed, one speed feature's tradeoff can shift as other features are integrated. While reverting a feature for coding gains, it is preferred to do it conservatively (i.e. With a much better tradeoff than the current bar). As Yeqing mentioned, it would be great if the speed loss can be reduced. 

@yunqingwang1 and @yeqing-wu 

If you are worried about encode time for speed 1,  note that my related PR was already merged, which speeds up the encoder by 8% (with some loss): https://github.com/AOMediaCodec/avm/pull/5403

So, if you combine #5403 and this PR,  we overall have a coding gain without any slowdown  (very small speed-up):

```
+---------+--------+-------------+
| Summary |  YUV   |   Enc-time  |
+---------+--------+-------------+
|   A1    | -0.06% |   100.10%   |
|   A2    | -0.19% |    99.94%   |
|Avg wo B2| -0.17% |    99.65%   |
+---------+--------+-------------+
```

I'd say that this is a pretty good trade-off, especially for speed 1 where quality is also important.
For faster speed settings, we can of course target faster encode times.
What do you think?

> Maybe only allow unrestricted H partitioning for low pyramid level frames? Or make it a resolution dependent feature? Please correct me if I miss anything.

I'll still go ahead and try this suggestion in the meantime.

yunqingwang1:
Ok. Since the combined PRs caused no speed regressions, I am good with this PR. Any other opinions?

yeqing-wu (approved):

PR #5399, opened 2026-09-11
Disable trellis for the dry pass

chengchen-google:
Performance on RA, speed 6:

```
Set 	BD-Rate (YUV)        Speedup		Ratio
A1     	0.25%	     	      2.47%	       	9.9
A2      0.26%                 2.67%         10.3
```

STATS_CHANGED

Change-Id: I40663e69ad96651e18191130bb333e48c65a026e


yeqing-wu:
The ratio of this PR is not good, A2 is around the boundary, A1 is missing the minimum ratio bar. Can we improve this one to achieve better tradeoff?

PR #5401, opened 2026-09-11
Replace fnv-1a with crc32 hashing

jingninghan:
Performance change:
    
    sp1: 0.02% 98.29%
    
    sp2: +0.04% 97.79%
    
    sp3: +0.02%  97.16%
    
    sp4: +0.01% 98.37%
    
    Thanks to linzhen@google.com for evaluating the performance.


gjin-rgb:
Do you have FG16 results for cpu2 and above?  It seems the QP restriction in CPU1 is also removed.

jingninghan:
> Do you have FG16 results for cpu2 and above? It seems the QP restriction in CPU1 is also removed.

Please see the updated result numbers.

Yes, i was hoping to remove its QP dependency if possible, as it would be helpful when the rate control kicks in later on.

yeqing-wu (approved):
LGTM

PR #5404, opened 2026-09-14
Fixes for 1-pass CBR rate control for realtime mode

marpan-max:
Update default buffer sizes and overshoot/undershoot tolerances for AVM_USAGE_REALTIME to low-latency settings. Tune CBR rate control logic:
- Limit upward Q transitions when buffer level is healthy
- Dampen active worst quality swings and adjust active best quality
- Scale frame 0 I-frame target to match the 600ms initial buffer

This improves the rate-targeting at low bitrates, and gives ~2% bdrate gain in rtc sets, for end-usage=cbr.


av2/encoder/ratectrl.c
```diff
@@ -302,16 +302,12 @@ int av2_rc_get_default_max_gf_interval(double framerate, int min_gf_interval) {
 void av2_rc_init(const AV2EncoderConfig *oxcf, int pass, RATE_CONTROL *rc) {
   const RateControlCfg *const rc_cfg = &oxcf->rc_cfg;
   int i;
+  (void)pass;
```
jianj-g: It seems like this is not used any more in this function so it can be removed from the argument list?

av2/encoder/ratectrl.c
```diff
@@ -421,6 +417,9 @@ static int adjust_q_cbr(const AV2_COMP *cpi, int q, int active_worst_quality) {
     }
     // Limit the decrease in Q from previous frame.
     if (rc->q_1_frame - q > max_delta) q = rc->q_1_frame - max_delta;
+    if (rc->buffer_level > (rc->optimal_buffer_level >> 3)) {
+      if (q - rc->q_1_frame > max_delta) q = rc->q_1_frame + max_delta;
```
jianj-g: you can merge the condition on line 421 and 420 into one if statement

av2/encoder/ratectrl.c
marpan-max: Done

av2/encoder/ratectrl.c
marpan-max: Done

jianj-g (approved):

